PR #6, opened 2022-10-14
COR-510 theming skeleton

neb42:
* Add package to manage theming

Needs to have CI/CD setup to generate theme before deploy. I went for this approach rather than committing the theme so that the sass and css files don't get out of sync.


web/templates/layout.gohtml
```diff
@@ -12,6 +12,7 @@
         <link rel="stylesheet" href="/static/bootstrap.min.css">
```
ludydoo, line 12: I think we can remove this whole line since the generated css will contain the themed everything
```suggestion
```

web/theme/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "theme",
+  "scripts": {
+    "build": "sass theme.scss:../static/theme.min.css --style=compressed"
```
ludydoo, line 4: maybe add an entry in the Makefile 
```
.PHONY: css
css:
  @cd web/theme/package.json && npx build
```

ludydoo, line 4: Also perhaps copy over the compressed javascript `bootstrap.bundle.min.js` (and maybe the bootstrap icons as well)

neb42: I can't see the icon files in the node modules

neb42:
Need to add bootstrap build to dockerfile, but depends on another pr

Makefile
```diff
@@ -18,4 +18,8 @@ proxy:
 
 .PHONY: serve
 serve:
-	@go run . serve --listen-address=:8080 --db-driver=postgres --db-dsn=postgres://postgres:postgres@localhost:5432/core?sslmode=disable --log-level=debug
\ No newline at end of file
+	@go run . serve --listen-address=:8080 --db-driver=postgres --db-dsn=postgres://postgres:postgres@localhost:5432/core?sslmode=disable --log-level=debug
+
+.PHONY: bootstrap 
+bootstrap:
+	@cd web/theme && yarn build
```
ludydoo, line 25: should we `yarn` before `yarn build` ?

ludydoo:
looking good to me, I think we have to merge ourselves now

neb42:
I've broke something

neb42:
Fixed

PR #8, opened 2022-10-17
COR-508 Enhance dockerfile with multi-stage builds

ludydoo:
Setting up the cicd pipelines will require a multi-stage build. Why: 
The first stage is based off a large golang:1.19 base image with a lot of libraries installed. We can use this "stage" as the "devel" stage, which contains tooling for testing, building, etc. 
But we don't want to ship a large container image with unnecessary stuff.  That's why the last "final" stage is a "distroless" container, a very lean container image with almost nothing on it. It's more secure, more lightweight. We simply copy the built executable from the first stage onto that final stage. 
The pipeline can still use the "devel" stage to launch tests from the built container image itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,4 +105,8 @@ certs/*
 
 generated.csv
 
+# Apple
 .DS_Store
+
+/coverage
+/reports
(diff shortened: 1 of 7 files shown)


internetti (commented):
can we say `dev` instead of `devel`? otherwise, all the make commands seem to work, I'm not able to judge the correctness of the Dockerfile

internetti (approved):

PR #14, opened 2022-10-19
COR-510: nrc styling

internetti:


internal/handlers/index.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 func HandleIndex(templates map[string]*template.Template) http.Handler {
```
ludydoo, line 8: Maybe rename to `HandleHome` ?

web/templates/layout.gohtml
```diff
@@ -1,32 +1,10 @@
...
-        <script src="/static/bootstrap.bundle.min.js"></script>
-
-        {{template "head" .}}
-    </head>
-    <body>
-    {{template "body" .}}
-    </body>
-{{end}}
-
 {{define "nav"}}
```
ludydoo, line 1: If layout only contains `nav`, could we rename it to `nav.gohtml` or `shared.gohtml` ?

ludydoo:
@internetti looks go to merge!

PR #16, opened 2022-10-19
COR-514-excel-upload

neb42:
* Support excel and csv uploads in handler
* Add individual excel unmarshal
* Map file column names to db column names (currently using the same names, but allows us to differ)


ludydoo:
Do you want to add tests in another PR ?

internetti (commented):
when trying to upload the bla.xlsx I get this error: `failed to put records: sql: Scan error on column index 15, name "country_id": converting NULL to string is unsupported`
will this be fixed later?

internal/api/individual_tabular.go
```diff
@@ -2,43 +2,78 @@ package api
...
+	f, err := excelize.OpenReader(reader)
+
+	if err != nil {
+		return err
+	}
+
+	// defer func() {
+	// 	if err := f.Close(); err != nil {
+	// 	}
+	// }()
```
ludydoo, line 36: Unused ?

internal/api/individual_tabular.go
```diff
@@ -2,43 +2,78 @@ package api
...
 	csvReader.TrimLeadingSpace = true
 	records, err := csvReader.ReadAll()
 	if err != nil {
-		return nil, err
+		return err
+	}
+	return unmarshalIndividualsTabularData(records, individuals, fields)
+}
+
+func UnmarshalIndividualsExcel(reader io.Reader, individuals *[]*Individual, fields *[]string) error {
```
ludydoo, line 26: I don't think you need the `fields` parameter here.

neb42:
> when trying to upload the bla.xlsx I get this error: `failed to put records: sql: Scan error on column index 15, name "country_id": converting NULL to string is unsupported` will this be fixed later?

This is the country id issue I mentioned in standup today. We used to upload it as part of the excel file, but now it is done via the query string. It was broke in a different pr and I didn't think it was part of this pr to fix it.

internal/api/individual_tabular.go
neb42: I think I do. This is using the same partial upload logic that you added for csv uploads

internal/constants/individual.go
```diff
@@ -44,3 +60,75 @@ const (
 	FileColumnIndividualPresentsProtectionConcerns = "presents_protection_concerns"
 	FileColumnIndividualSensoryImpairment          = "sensory_impairment"
 )
+
+var IndividualDBColumns = []string{
```
ludydoo, line 64: Unused variable ?

internal/constants/individual.go
```diff
@@ -28,6 +28,22 @@ const (
...
+	DBColumnIndividualFullName                   = "full_name"
+	DBColumnIndividualGender                     = "gender"
+	DBColumnIndividualIsMinor                    = "is_minor"
+	DBColumnIndividualMentalImpairment           = "mental_impairment"
+	DBColumnIndividualPhoneNumber                = "phone_number"
+	DBColumnIndividualPhysicalImpairment         = "physical_impairment"
+	DBColumnIndividualPreferredName              = "preferred_name"
+	DBColumnIndividualPresentsProtectionConcerns = "presents_protection_concerns"
+	DBColumnIndividualSensoryImpairment          = "sensory_impairment"
+
```
ludydoo, line 46: Not sure about the `DBColumn...` name, since it's not used in the DB at all. But we can check it later

neb42:
I'm not going to add tests for `UnmarshalIndividualsExcel` and `UnmarshalIndividualsCSV` as they are fairly simple, harder to test, and could be better handled by some integration tests once we're ready.

PR #18, opened 2022-10-20
COR-508 proper jwt authentication

ludydoo:
## First part for supporting refresh tokens

Azure Authentication does not refresh the id/access tokens automagically. So at some point, users will get a "Forbidden" page without being automatically redirected to login again. So we must include a mechanism in the application to detect when the session is about to expire. 

The incoming jwt token contains an `exp` claim, which corresponds to the expiration unix timestamp of the token. We can rely on this information to implement our mechanism. For example, if the token expires in 6 hours, perform a refresh.

This is something that the `browser` must do. Azure explains how to perform the refresh:  
See https://learn.microsoft.com/en-us/azure/app-service/configure-authentication-oauth-tokens#refresh-auth-tokens for details. 

This PR removes support for "unverified" claims, and implements proper jwt validation, fetching of jwks, and so on. We can then safely access the `exp` claim and build on that. 

Further PR will be issued to actually implement the refresh mechanism.


internetti (commented):
where is the fetching of the jwks happening? is that happening internally, by this oicd library?

ludydoo:
@internetti yes, the oidc library takes care of fetching (and refreshing) the jwks

internetti (approved):

PR #33, opened 2022-10-24
COR-504 fix token refresh

ludydoo:


web/templates/base.gohtml
```diff
@@ -20,27 +20,49 @@
...
+
+    function delay(ms) {
+        return new Promise(resolve => setTimeout(resolve, ms));
+    }
+
+    async function retryPromise(fn, tentatives, delayMs) {
+        const errors = [];
+        while (errors.length < tentatives) {
+            try {
+                return await fn();
```
neb42, line 34: I think this should just be: `return fn();`

web/templates/base.gohtml
```diff
@@ -20,27 +20,49 @@
...
-                .then(() => fetch("{{ $refreshURL }}"))
-                .then(() => pollRefreshToken())
+
+<script type="application/javascript">
+
+    function delay(ms) {
+        return new Promise(resolve => setTimeout(resolve, ms));
+    }
+
+    async function retryPromise(fn, tentatives, delayMs) {
```
neb42, line 30: Maybe rename `tentatives` to `maxErrorCount` or something

web/templates/base.gohtml
```diff
@@ -20,27 +20,49 @@
...
-{{end}}
+        throw new Error("Max tentatives reached");
+    }
+
+    async function pollRefreshToken() {
+        while (true) {
+            try {
+                await retryPromise(async () => {
+                    const resp = await fetch("{{tokenRefreshURL}}")
+                    if (resp.status > 399) {
```
neb42, line 48: `>= 400` is more readable I think

web/templates/base.gohtml
```diff
@@ -20,27 +20,49 @@
...
+                window.location.href = "{{loginURL}}";
+            }
+        }
+    }
+
+    (async () => {
+        await pollRefreshToken();
+    })();
+
+    pollRefreshToken()
```
neb42, line 64: Why are you calling it twice? The second seems sufficient

neb42:
lgtm

PR #41, opened 2022-10-24
Change upload to be a single action and disable while uploading

neb42:


web/templates/individuals.gohtml
```diff
@@ -27,6 +27,16 @@
                     }
                 }
             }
+
+            function openDialog() {
+                document.getElementById('individual-upload--file').click();
+            }
```
ludydoo, line 33: Unused?

neb42: No it's used a few lines below:
```
document.getElementById('individual-upload--button').addEventListener('click', openDialog);
```

neb42:
The tests added here are on the wrong branch 🤦

internal/auth/interface.go
```diff
@@ -69,3 +69,7 @@ func (p permissions) GetCountryIDsWithReadWritePermissions() containers.StringSe
 func (p permissions) hasExplicitReadWritePermissionInCountry(countryID string) bool {
 	return p.allowedCountryIDs.Contains(countryID)
 }
+
+func (p permissions) countryExists(countryID string) bool {
+	return p.allCountryIDs.Contains(countryID)
+}
```
ludydoo, line 75: I think this is unused

internal/server/middleware/has_country_permission_test.go
```diff
@@ -0,0 +1,65 @@
...
+func TestHasCountryPermission(t *testing.T) {
+	var permissions = []struct {
+		permission auth.Permission
+	}{
+		{permission: auth.PermissionRead},
+		{permission: auth.PermissionWrite},
+	}
+
+	var parametrizedTests = []struct {
+		allowedCountryIDs containers.Set[string]
```
ludydoo, line 22: add a `name` field to the struct. This will be more helpful when dealing with failures to identify the failing test. Go has native support for nested test cases. It's similar to `describe(..., () => it("name", ()=>{}))`

neb42: Nice, yeah debugging the tests was a nightmare

PR #42, opened 2022-10-25
COR-529 field validation errors

ludydoo:


internal/api/meta/meta.go
```diff
@@ -0,0 +1,53 @@
+package meta
```
neb42, line 1: Again, something to live in a shared library. Also seems weird that this doesn't already exist in some form.

internal/api/validation/individual.go
```diff
@@ -0,0 +1,92 @@
...
+	if birthDate != nil {
+		if birthDate.After(time.Now()) {
+			allErrs = append(allErrs, validation.Invalid(path, birthDate, "birthdate cannot be in the future"))
+		}
+		if *birthDate == (time.Time{}) {
+			allErrs = append(allErrs, validation.Invalid(path, birthDate, "must be a valid date"))
+		}
+	}
+	return allErrs
+}
```
neb42, line 47: Currently when parsing the excel, if the birth date is in an invalid format it will error and fail. This means we fail at the first instance and won't reach this validation. 

My first thought was to set the field to `nil` when parsing an error and raise at the validation stage. But if `nil` is an allowed value then we're hiding errors in the upload.

Maybe this needs to be done cell-by-cell in the unmarshelling?

neb42 (commented):
I think everything in:
```
api/errors/
api/meta/
validation/
```
should be in a common library for dealing with apis. This is also the sort of thing that I would look for an existing library to help with.

I don't think it's worth making a new repo now, but maybe this could all be moved to the same package? This would make it easier to factor out later.

internal/api/validation/individual.go
ludydoo: We must consider unmarshaling and validation as 2 different things. The unmarshaling will decode without validating. So we can have multiple formats (json,csv,excel) and a single validation logic. We can use a similar approach with appending to an `allErrs` list when we unmarshal, though in this case the errors will be those of type conversion errors rather than "business logic" validations

internal/api/errors/errors.go
```diff
@@ -0,0 +1,369 @@
+package errors
```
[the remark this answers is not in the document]
ludydoo: Agreed

internal/api/meta/meta.go
ludydoo: Yeah, this is a very useful mechanism for dealing with common error formats. It would make sense to have this in a shared library. But this is kind of out of scope for this one.. 
Having a typed error is really useful, especially when dealing with validation errors, and how to find the actual path/field of the error, as well as asserting the error type

internal/api/validation/individual.go
ludydoo: @neb42 I have a local spike on how to validate the `ListIndividualsOptions` aka `GetAllOptions` (should be renamed) that actually does this. Instead of failing early, it will append all encountered errors to a list and return that

neb42:
> @neb42 I would suggest we introduce (perhaps in a future PR) two different api subpackages
> 
> ```
> /api
>   /meta
>   /core  
> ```
> 
> That would prevent us from the overhead of having an external repo for now, and it would be easier to discern between the `core` types and the more "fundamental" or "meta" types, such as `Status`, `StatusError`, etc.

What about:
```
pkg/
  api/
internal/
  api/
```
Maybe with a different name to `api` to prevent conflicts. This would save us from moving all the existing code in `api` and my understanding is that `pkg` is for external code and `internal` for internal code.

ludydoo:
> 

we wouldnt have conflicts
e.g.

```
/api
  /core
    individual.go
    country.go
  /meta
    errors.go
    status.go
```

```
import (
  "github.com/NorwegianRefugeeCouncil/core/internal/api/core"
  "github.com/NorwegianRefugeeCouncil/core/internal/api/meta"
)

func myFunc(i core.Individual) error {}
func myFunc2(a meta.Status) error {}
```

neb42:
lgtm

PR #53, opened 2022-10-28
COR-533-add-batch-delete-action

neb42:


internal/db/individual.go
```diff
@@ -322,6 +324,29 @@ func (i individualRepo) softDeleteInternal(ctx context.Context, tx *sqlx.Tx, id
 	return nil
 }
 
+func (i individualRepo) SoftDeleteMany(ctx context.Context, ids []string, countryId string) error {
```
ludydoo, line 327: No need to pass countryId in this case

internal/db/individual.go
```diff
@@ -303,7 +305,7 @@ func (i individualRepo) putInternal(ctx context.Context, tx *sqlx.Tx, individual
 func (i individualRepo) SoftDelete(ctx context.Context, id string) error {
 	_, err := doInTransaction(ctx, i.db, func(ctx context.Context, tx *sqlx.Tx) (interface{}, error) {
 		err := i.softDeleteInternal(ctx, tx, id)
-		return err, nil
+		return nil, err
```
ludydoo, line 308: Convert to using `i.softDeleteInternalMany` and remove `i.softDeleteInternal`

internal/db/individual.go
```diff
@@ -322,6 +324,29 @@ func (i individualRepo) softDeleteInternal(ctx context.Context, tx *sqlx.Tx, id
...
+	})
+	return err
+}
+
+func (i individualRepo) softDeleteManyInternal(ctx context.Context, tx *sqlx.Tx, ids []string, countryId string) error {
+	l := logging.NewLogger(ctx).With(zap.Strings("individual_ids", ids))
+	l.Debug("deleting individuals")
+
+	const query = "UPDATE individuals SET deleted_at = $1 WHERE id = ANY($2) and deleted_at IS NULL and country_id = $3"
```
ludydoo, line 339: `WHERE id IN ($2)`

internal/db/individual.go
```diff
@@ -322,6 +324,29 @@ func (i individualRepo) softDeleteInternal(ctx context.Context, tx *sqlx.Tx, id
...
+func (i individualRepo) SoftDeleteMany(ctx context.Context, ids []string, countryId string) error {
+	_, err := doInTransaction(ctx, i.db, func(ctx context.Context, tx *sqlx.Tx) (interface{}, error) {
+		err := i.softDeleteManyInternal(ctx, tx, ids, countryId)
+		return nil, err
+	})
+	return err
+}
+
+func (i individualRepo) softDeleteManyInternal(ctx context.Context, tx *sqlx.Tx, ids []string, countryId string) error {
+	l := logging.NewLogger(ctx).With(zap.Strings("individual_ids", ids))
```
ludydoo, line 336:
```
idSet := containers.NewStringSet(ids...)
ids = idSet.Items()
```

internal/db/individual.go
```diff
@@ -322,6 +324,29 @@ func (i individualRepo) softDeleteInternal(ctx context.Context, tx *sqlx.Tx, id
...
+
+	const query = "UPDATE individuals SET deleted_at = $1 WHERE id = ANY($2) and deleted_at IS NULL and country_id = $3"
+	var args = []interface{}{time.Now().UTC(), pq.Array(ids), countryId}
+
+	if _, err := tx.ExecContext(ctx, query, args...); err != nil {
+		l.Error("failed to delete individuals", zap.Error(err))
+		return err
+	}
+
+	return nil
```
ludydoo, line 347: Check if the `RowsAffected` in the `ExecContext` result = `len(ids)`, otherwise throw an error

internal/db/individual.go
neb42: Same for this endpoint except the individual ids are in the payload

web/templates/individuals.gohtml
```diff
@@ -59,6 +59,9 @@
 
                 <div class="d-flex justify-content-end align-items-center">
                     {{ if or (.RequestContext.Auth.IsGlobalAdmin) (.RequestContext.Auth.HasCountryLevelPermission .RequestContext.SelectedCountryID 1)}}
+                        <button type="button" class="btn btn-outline-primary me-2" onclick="submitActionForm('delete')">
+                            Batch delete 
+                        </button>
```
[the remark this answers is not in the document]
neb42: Yeah, I'll open a separate ticket for UI/UX. I just threw a button in for now

internal/db/individual.go
neb42: Is there any difference?

internal/db/individual.go
ludydoo: I think for this specific usecase, we should 
- modify the `ListIndividualOptions` to be able to pass multiple `IDs`. 
- in the deleteMany handler, first retrieve the individuals about to be deleted by IDs
- check if all individuals.CountryID = selectedCountryID. Otherwise throw
- perform the delete.

neb42: > CountryID is more of a web-presentation filter

I disagree on this. We are restricting access to what users can see. A user in one country should not be able to see the data of a user in another country. Currently they can (it may not be easy, but they can).

> It's basically setting the intent that we consider the country id to be part of the key of the individual, so {ID, CountryID} tuple rather than {ID}

I don't think the key needs to change

> You're suggesting that adding this in the DB layer would prevent from programming mistakes 

No, my concern isn't with programming mistakes. My concern is with malicious users accessing data they do not have access to. Although more safety against programming mistakes is good. A bad actor with some compute could do:
```
countryId = "country-id-with-access"
while (True) {
  id = uuidv4() // Chances of them hitting the correct id are slim, but maybe they have a list of ids
  response = get("https://core.com/countries/${countryId}/individuals/${id}")
  if (response.status == 200) {
    saveResponse(response);
  }
}
```

In the future, yes there may be some valid use cases for querying individuals without country id, but for now I think we need to check this somewhere. And not only for this endpoint.

I would normally have multiple db methods for different use cases, such as:
```
GetIndividualByIdInCountryId
GetIndividualByPhoneNumber
```
Maybe they call the same thing internally.

ludydoo: > We are restricting access to what users can see. 

Exactly. This is a **presentation concern**. (What users can see). 

**What users see** is a subset of **what users can see** only because we made a **design decision** to confine them to 1 country at a time in one discrete server-rendered web application. But this change leaks this design decision into the persistence layer where it does not belong. 

Pushing this logic to the absurd limit:  `Do not allow users to delete individuals that do not have a corresponding <tr>{{individual.ID}}</tr> element rendered on the web page`. Of course it's absurd, but there is a common theme with saying `Do not allow users to delete individuals whose countryID don't match the https://core.nrc.no/countries/<country_id>/ present in the POST request`.

Maybe what is confusing is that the way that Core is currently written does not distinguish between the web application and the resource server, as they are both mingled together right now. Typically, the web application would be a separate runtime than the one of the resource server. So these presentation concerns (selected country) would be only be in the web application and not the resource server. 

If we would write an SPA for rendering the website, these presentation concerns would be isolated in the SPA, and the concept of "Selected Country" would not appear anywhere in the backend. In its current format, these presentation concerns should go no further than the handler. A good API would only receive an individual ID and deal with it. It effectively does not need the CountryID at all to locate an individual and perform all its duties. The reason we include it in the path right now is just because we are writing a server-side rendered app and it's useful for navigation. 

> A user in one country should not be able to see the data of a user in another country. Currently they can (it may not be easy, but they can).

This is just bad logic on the application side. This doesn't need to leak to the persistence layer. 

> I don't think the key needs to change

They would effectively change. If you need both {CountryID,ID} to access any individual, this is the effective key.

Also, we are aiming for generic forms. We will not know at compile time whatever scope a generic form is confined to. So the persistence layer should remain as generic as possible, and possibly follow the CRUD/REST pattern. It's currently not generic, but this would make it even further specialized into dealing with the "individual" type, rather than a generic entity type.

neb42: > This is a presentation concern

It's not. Our authorisation policy is based on countries. Currently users can access resources in countries that they are not authorised to access.

> This doesn't need to leak to the persistence layer.

I think it should for safety, but I'm happy to concede that. We need to at least be validating that a user can perform these actions though and we currently aren't.

I don't think the key needs to change. Using the url `/country/{country_id}/individual/{individual_id}` makes things a bit simpler for us, but we still need to validate that they're allowed to access that resource. We could do `/individual/{individual_id}` with an additional query.

ludydoo: > It's not. Our authorisation policy is based on countries. Currently users can access resources in countries that they are not authorised to access.

In this context, it is. The path parameter corresponds to the selected country ID. But as you mention, there can be a mismatch between the individual.CountryID and the selected Country ID. This should be handled in the handler, and we should not rely on the database to perform our authorization policy. 

```
ind := repo.GetIndividual(ctx, individualID)
if ind.CountryID != selectedCountryID {
  http.Error(w, http.StatusNotFound, "some useful error")
  return
}
```

which can be generalized for the batch actions as well (list/delete/update)

```go
// ###########
// delete handler
// ###########

individualIdsToDeleteSlice := strings.Split(r.Form["individualIDs"], ",")
if len(individualIdsToDeleteSlice) == 0 {
  http.Error(w, http.StatusBadRequest, "no individual ids provided")
}
individualIdsToDelete := containers.NewStringSet(individualIdsToDeleteSlice...)

individualsToDelete, err := repo.ListIndividuals(ctx, GetAllOptions{IDs: individualIdsToDelete})
// could also be
individualsToDelete, err := repo.ListIndividuals(ctx, GetAllOptions{IDs: individualIdsToDelete, CountryID: selectedCountryID})
if err != nil {
  http.Error(w, http.StatusInternalServerError, "internal error")
  return
}

foundIndividualIds := containers.NewStringSet()
for _, foundIndividual := range individualsToDelete {
  foundIndividualIds.Add(foundIndividual.ID)
}

// if not passing the countryID to the repo.List
for _, i := range individualsToDelete {
  if i.CountryID != selectedCountryID {
    http.Error(w, http.StatusNotFound, fmt.Sprintf("individual %s was not found in country %s", i.ID, selectedCountryID))
    return
  }
}

// if passing the country ID to the repo.List
for _, i := range individualsToDelete {
  if !foundIndividualIds.Contains(i.ID) {
    http.Error(w, http.StatusNotFound, fmt.Sprintf("individual %s was not found in country %s", i.ID, selectedCountryID)
  }
}

// at this point all the individuals specified in the Delete request are actually in the selected country, safe to delete. 
repo.DeleteMany(ctx, individualIdsToDelete)


// ########
// put handler
// ########

individualIDsToUpdate := containers.NewStringSet()
for _, newIndividual := range individuals {
  individualIDsToUpdate.Add(newIndividual.ID)
}
oldIndividuals := repo.ListIndividuals(ctx, GetAllOptions{IDs: individualIDsToUpdate})
for _, i := range oldIndividuals {
  if i.CountryID != selectedCountryID {
    http.Error(w, http.StatusConflict, fmt.Sprintf("An individual with ID %s already exists in another country", i.ID))
    return
  }
}
// We are not updating individuals from other countries than the selected country at this point
repo.PutMany(ctx, individuals)

// #########
// list handler
// #########

opts := parseOptionsFromQueryParams()
opts.CountryID = selectedCountryID // this ensures that only individuals in the selected country are returned
repo.List(ctx, opts)
```

internal/db/individual.go
```diff
@@ -20,9 +21,11 @@ import (
 type IndividualRepo interface {
 	GetAll(ctx context.Context, options api.GetAllOptions) ([]*api.Individual, error)
 	GetByID(ctx context.Context, id string) (*api.Individual, error)
+	ListByID(ctx context.Context, ids []string) ([]*api.Individual, error)
```
ludydoo, line 24: If we would add the `IDs` parameter in the `GetAllOptions` we would be able to solve partially the ability for a user to search by mutliple IDs

internal/handlers/individuals_delete.go
```diff
@@ -0,0 +1,64 @@
...
+		for _, individual := range individuals {
+			if individual.CountryID != countryID {
+				l.Error("individual does not belong to selected country", zap.String("individual_id", individual.ID), zap.String("country_id", countryID))
+				http.Error(w, fmt.Sprintf("individual %s does not belong to selected country", individual.ID), http.StatusBadRequest)
+				return
+			}
+		}
+
+		if err := repo.SoftDeleteMany(ctx, individualIds); err != nil {
+			l.Error("failed to delete individual", zap.Error(err))
```
ludydoo, line 57: failed to delete individual**s**

internal/db/individual.go
```diff
@@ -195,6 +198,33 @@ func (i individualRepo) getByIdInternal(ctx context.Context, tx *sqlx.Tx, id str
...
+		return ret, nil
+	}
+
+	const query = "SELECT * FROM individuals WHERE id IN ($1) and deleted_at IS NULL"
+	var args = []interface{}{pq.Array(ids)}
+	err := tx.SelectContext(ctx, &ret, query, args...)
+	if err != nil {
+		return nil, err
+	}
+	return ret, nil
```
ludydoo, line 225: Consider: add batching logic to not go above the max parameter limit for postgres. See https://github.com/NorwegianRefugeeCouncil/core/pull/63

neb42: I'll do once your batch pr is merged

ludydoo: Done

neb42: actually not relevant if moving ids to getalloptions

ludydoo:
lgtm

ludydoo (approved):

PR #54, opened 2022-10-29
UI Tweaks

ludydoo:


web/templates/individuals.gohtml
```diff
@@ -6,7 +6,7 @@
         function goToIndividual() {
             const id = document.getElementById("goToIndividualId").value;
             if (id !== "") {
-                window.location.href = "{{.BaseURL}}{{.CurrentUrl}}/" + id;
+                window.location.href = "/countries/{{.Options.CountryID}}/individuals/" + id;
```
neb42, line 9: I think `.Options.CountryID` isn't the right place to get this from. This made sense when country id was part of the filters, but now we have a single country id it shouldn't be in the options. I think `.RequestContext.SelectedCountryID` is better.

web/templates/individuals.gohtml
```diff
@@ -267,7 +308,7 @@
                 <div class="col-12">
                     <div class="mb-5">
                         <div class="overflow-scroll mb-3">
-                            <table class="table">
+                            <table class="table table-responsive" style="white-space: nowrap">
```
neb42, line 311: `table-responsive` should be on the parent div. If you rebase on main I've already added it in

web/templates/individuals.gohtml
ludydoo: I see what you mean. 

If we change this to `.RequestContext.SelectedCountryID`, then we have inconsistency in the way we deal with the "selected country scope". For example, should we also do the same for the `individual` and `country` views ? They both require the concept of a "selected country", yet we do not use the `.RequestContext.SelectedCountryID`, but rather the `.Country.ID` or `.Individual.CountryID`. 

Idea: Perhaps it would help the readability to rename `.Options` to `.IndividualListOptions`.

My rationale is that I would like for each view being dedicated to managing a single entity type:
```
individuals.gohtml allows the user to easily manage the IndividualListOptions entity type
individual.gohtml allows the user to easily manage the Individual entity type
countries.gohtml allows the user to easily manage the CountryListOptions entity type
country.gohtml allows the user to easily manage the Country entity type
```
In each case, we have 3 strategies for parsing an entity:
1. Obtain a property from the `PathParams`, e.g. For a `Country`, we parse the `Country.ID` from the path param `/countries/{country_id}` 
2. Obtain a property from the `QueryParam`. e.g. For `IndividualListOptions`, we obtain `take`, `skip`, etc from the query params: `/countries/{country_id}/individuals?skip=<skip>&take=<take>&...
3. Obtain a property from the `PostForm`. e.g. for `Individual`, we obtain the `FullName`, `Address`, ... from the POST parameters. Example:
```
POST /countries/{country_id}/individuals/{individual_id} HTTP/1.1
Host: core.nrc.no
Content-Type: application/x-www-form-urlencoded

address=BlvdDrive%20100&full_name=John%20Doe
```

I think it would help consistency throughout the application if we (as much as possible) follow the strategy of 1 view = 1 entity type, as well as having all available properties of the Entity under a single key.

If we consider the `individuals.gohtml` view as a view for editing the `ListIndividualsOptions` entity (aka `GetAllOptions`), it would make sense to keep all the related properties under a single object.

ludydoo: bla

web/templates/individuals.gohtml
```diff
@@ -59,22 +59,22 @@
...
                             <input
                                 hidden
                                 id="individual-upload--file"
-                                name="file"
-                                type="file"
+                                name="file" 
+                                type="file" 
                                 class="form-control"
-                                accept="text/csv, application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+                                accept="text/csv, application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
```
neb42, line 77: Something needs fixing in your editor setup

web/templates/individuals.gohtml
```diff
@@ -274,8 +315,8 @@
 
                 <div class="col-12">
                     <div class="mb-5">
-                        <form id="individuals-action-form" class="table-responsive mb-3" method="post">
-                            <table class="table">
+                        <form id="individuals-action-form" class="overflow-scroll table-responsive mb-3" method="post">
+                            <table class="table" style="white-space: nowrap">
```
neb42, line 319: I'd just remove this. Irena has a pr open that has the table scrolling nicely.

web/templates/individuals.gohtml
ludydoo: strangwe

neb42:
lgtm

PR #58, opened 2022-11-01
COR-535: make first two columns sticky and table scroll horizontally

internetti:
chose a hybrid solution between option 1 and 2 in the ticket

table scrolls horizontally, first two columns are sticky (the select box, and the name column, cause it was the first one currently)

### DIFF
--- a/web/templates/individuals.gohtml
+++ b/web/templates/individuals.gohtml
@@ -105,10 +105,10 @@
                                 <div class="form-group mb-3 col-6">
                                     <label class="form-label" for="ID">IDs</label>
                                     <input id="ID"
-                                            name="id"
-                                            type="text"
-                                            placeholder="Search by ID"
-                                            class="form-control"
+                                           name="id"
+                                           type="text"
+                                           placeholder="Search by ID"
+                                           class="form-control"
                                            value="{{if .Options.IDs}}{{joinStrings .Options.IDs ","}}{{end}}">
                                 </div>
                                 <!-- End ID -->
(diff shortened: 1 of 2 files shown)


neb42:
I think the last sticky column could do with a border to provide a bit of separation

internetti:
@neb42 is the light gray background not enough as a separating feature?

ludydoo:
@neb42 @internetti is that good to merge ?

internetti:
that depends on wether the grey background of the sticky columns is enough of a separator, or if it really needs a border.
but if you need this merged urgently, then merge it, we can figure out the border later

neb42:
I think it needs the border. It looks a bit weird currently. The sticky cells blend into then non-sticky ones. Happy to merge and do it in another pr if it's blocking

neb42:
And maybe remove the row hover colour?

neb42:
There's three very similar shades of grey going on here

image

internetti:
@ludydoo I'm having a bit of trouble with this, the border is moving for some reason, so if you want to wait, it seems to still take a while. otherwise, merge it, I'll fix it in another PR

internetti:
sticky elements in general seem to have issues with borders. here's a suggestion. couldn't make the header have a more visible border color. 
off for today, feeling quite bad

neb42:
Yeah it'd be nice if the header border was a bit more visible, but I think this looks great otherwise. lgtm

web/templates/individuals.gohtml
```diff
@@ -322,15 +346,38 @@
                                             </td>
                                             <td>
                                                 {{if .Email}}
-                                                    <a href="mailto:{{.Email}}">{{.Email}}</a>
```
ludydoo, line 325: Why remove the "mailto" ? This would allow users to click and open their mail editor directly to send a message to the individual. I think it's pretty useful. Perhaps this could be converted to a dropdown eventually.

neb42: I find the "mailto" and phone number incredibly annoying. It makes it harder to copy and paste as well. I am 100% in favour of removing it.

internetti: because the ticket said that only the id (currently the name, because we don't have the id yet) is clickable, the rest isn't. maybe this is a Fix decision

PR #64, opened 2022-11-02
COR-523 Implement sorting

ludydoo:


ludydoo:
There was a missing closing `<div>` in the `individuals.gohtml` hence the large diff

internal/db/individual.go
```diff
@@ -392,3 +419,28 @@ func (i individualRepo) softDeleteManyInternal(ctx context.Context, tx *sqlx.Tx,
...
+		return constants.DBColumnIndividualFullName, nil
+	case "phoneNumber":
+		return constants.DBColumnIndividualPhoneNumber, nil
+	case "normalizedPhoneNumber":
+		return constants.DBColumnIndividualNormalizedPhoneNumber, nil
+	case "age":
+		return constants.DBColumnIndividualBirthDate, nil
+	case "gender":
+		return constants.DBColumnIndividualGender, nil
+	case "email":
```
neb42, line 441: Don't we have constants we can use for these case strings?

internal/db/individual.go
```diff
@@ -192,7 +193,33 @@ func (i individualRepo) unbatchedGetAllInternal(ctx context.Context, tx *sqlx.Tx
...
+			case api.SortDirectionAscending:
+				query = query + " ASC"
+			case api.SortDirectionDescending:
+				query = query + " DESC"
+			default:
+				query = query + " ASC"
+			}
+		}
+	} else {
+		query = query + " ORDER BY created_at DESC"
```
neb42, line 220: I think there was a product requirement to order by id, but I think `created_at` is better. Maybe just something to raise with Fix

web/templates/individuals.gohtml
```diff
@@ -323,120 +412,172 @@
...
+    {{$maxWidth := .MaxWidth}}
+    {{$sortDirection := $options.GetSortDirectionForField $fieldName}}
+    {{$ascSortUrl := $options.GetSortURLForField $fieldName "ascending"}}
+    {{$descSortUrl := $options.GetSortURLForField $fieldName "descending"}}
+
+    {{ if not $sortable}}
+        <th style="min-width: 5rem">
+            <a class="d-inline-block text-decoration-none text-black">
+                {{$label}}
+            </a>
```
neb42, line 549: Why an anchor tag in the column header?

neb42 (commented):
On hover the sort icon should be orange

internal/db/individual.go
ludydoo: Yeah I've set up a pr to generate the `json` tag constants, we could use that

internal/db/individual.go
ludydoo: Because sorting by UUID.. does not make much sense

web/templates/individuals.gohtml
ludydoo: whoops. Corrected

neb42:
> On hover the sort icon should be orange

Still needs this

PR #79, opened 2022-11-04
COR-582 add constants for new fields

ludydoo:
Added constants representing the new fields for the individual. 
Some of the constants need deleting (Impairments). Will do once the individual is fully merged in. 
Renamed a few constants that were missing a "s" in the name. (`FormParam` vs `FormParams`)


neb42:
lgtm

PR #80, opened 2022-11-04
COR-583 rename GetAllOptions

ludydoo:
Renaming GetAllOptions to ListIndividualsOptions and putting it in its own file. Feels good


internal/handlers/individuals_download.go
```diff
@@ -46,7 +46,7 @@ func HandleDownload(
 			format = "xlsx"
 		}
 
-		ret, err := userRepo.GetAll(ctx, api.GetAllOptions{})
+		ret, err := userRepo.GetAll(ctx, api.ListIndividualsOptions{})
```
neb42, line 49: Not really to do with this pr, but this should be `individualsRepo`

neb42:
lgtm

PR #87, opened 2022-11-07
COR-538: next new individual

internetti:
add a "save & continue" button, that redirect to a new form after submitting

### DIFF
--- a/web/templates/individual.gohtml
+++ b/web/templates/individual.gohtml
@@ -7,18 +7,27 @@
         document.addEventListener("DOMContentLoaded", function () {
             const form = document.getElementById("individualForm")
             const submitButton = document.getElementById("submitButton")
+            const submitAndContinueButton = document.getElementById("submitAndContinueButton")
             const deleteButton = document.getElementById("deleteButton")
             const confirmDeleteButton = document.getElementById("confirmDeleteButton")
             submitButton.addEventListener("click", function () {
                 submitButton.disabled = true
                 deleteButton.disabled = true
+                submitAndContinueButton.disabled = true
                 confirmDeleteButton.disabled = true
                 form.submit()
+            })
+            submitAndContinueButton.addEventListener("click", function () {
+                form.submit()
+                setTimeout(function () {
+                    window.location.assign(window.location.protocol + "//" + window.location.host + "/countries/" + {{.RequestContext.SelectedCountryID}} + "/individuals/new")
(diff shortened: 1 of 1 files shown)


ludydoo:
I think that this should be handled by the backend. If we put a checkbox `<input type="checkbox" name="createAnotherIndividual"/>`,
then in the backend, instead of redirecting the user to the new individual page, we put
```
if r.Form.Get("createAnotherIndividual") == "true" {
  http.Redirect(w, r, http.StatusSeeOther, "/countries/.../individuals/new")
  return
}
```

web/templates/individual.gohtml
```diff
@@ -35,7 +44,7 @@
                 </h1>
                 <form id="individualForm"
                       method="post"
-                      action="/countries/{{.Individual.CountryID}}/individuals/{{if .Individual.ID}}{{.Individual.ID}}{{else}}new{{end}}">
+                      action="/countries/{{.RequestContext.SelectedCountryID}}/individuals/{{if .Individual.ID}}{{.Individual.ID}}{{else}}new{{end}}">
```
ludydoo, line 47: Why this change ?

internetti: because the former version did not work. it kept redirecting to the countries page, without submitting. I think Individual.CountryID is not defined before submit?

internetti:
regarding the redirecting, not against the backend approach per se, but we don't currently have validation

web/templates/individual.gohtml
ludydoo: It should be, it is set for both new individuals and individuals being updated before rendering the page

ludydoo:
> regarding the redirecting, not against the backend approach per se, but we don't currently have validation

yeah we do have validation, the `individual` handler has this:

```
		// Validate the individual
		validationErrors = apivalidation.ValidateIndividual(individual)
		if len(validationErrors) > 0 {
			render()
			return
		}
```

web/templates/individual.gohtml
internetti: okay, but it still didn't work, does it work for you with the former version?

internetti:
yeah, I didn't mean that validation is not implemented, just not wished for by product at the moment. anyway, will see what I can do about the redirection.

internetti:
I'll put this on hold, I'd say the product tickets have a higher priority, this is more a nice to have

PR #96, opened 2022-11-09
COR-523 sorting

ludydoo:
Adds sorting for the individual list view


internetti (commented):
I don't think it's necessary to implement sorting for multiple columns, the ticket sys only one column at a time, and the ui also does not allow for more than one

web/templates/individuals.gohtml
```diff
@@ -657,136 +665,435 @@
...
+                                    <div style="width: 6rem">
+                                        {{.Gender}}
+                                    </div>
                                 </td>
+
                                 <td>
-                                    {{if .Age}}
-                                        {{.Age}}
-                                    {{end}}
+                                    <div style="width: 4rem">
```
internetti, line 965: is it necessary to have the wrapping divs? adding the widths to the tds doesn't do anything?
also, could we use variables for these widths? checking manually if the widths here fit the header widths would not be ideal

ludydoo: Yeah it's necessary, you're right, setting the width on the `td` doesn't do anything. I'll add some variables for the widths

web/theme/theme.scss
```diff
@@ -326,12 +326,41 @@ a.home-link {
...
+    }
+
+    .sort-icon-current {
+        // show current sort icon in table header
+        display: inline-block;
+    }
+
+    .sort-icon-next {
+        // do not show next sort icon in table header
+        display: none;
```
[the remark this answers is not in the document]
ludydoo: Fixed

web/templates/individuals.gohtml
```diff
@@ -649,145 +658,519 @@
...
+                                    <div style="width: {{$freeField3Width}}">
+                                        {{.FreeField3}}
+                                    </div>
                                 </td>
+                                <!-- End FreeField3 -->
+
+                                <!-- FreeField4 -->
                                 <td>
-                                    {{.FreeField4}}
+                                    <div style="width: 8{{$freeField4Width}}">
```
internetti, line 1145:
```suggestion
                                    <div style="width: {{$freeField4Width}}">
```

web/templates/individuals.gohtml
```diff
@@ -649,145 +658,519 @@
...
 
+            {{$checkBoxWidth := "4rem"}}
+            {{$dateWidth := "8rem"}}
+            {{$dateTimeWidth := "10rem"}}
+            {{$identWidth := "8rem"}}
+
+            {{$selectWidth := "2rem"}}
+            {{$idWidth := "20rem"}}
+            {{$nameWidth := "10rem"}}
+            {{$genderWidth := "6rem"}}
```
internetti, line 669:
```suggestion
            {{$genderWidth := "8rem"}}
```
doesn't make too much sense to cut this off

internetti (approved):

PR #99, opened 2022-11-09
Validate max length

ludydoo:


ludydoo:
Yeah I had this in the big individual form PR. Will add it again

internetti (approved):

PR #102, opened 2022-11-09
COR-608: fix download

internetti:


internal/api/individual_tabular.go
```diff
@@ -316,25 +316,60 @@ func MarshalIndividualsExcel(w io.Writer, individuals []*Individual) error {
...
-			row[j] = strconv.FormatBool(value.(bool))
+		case DisabilityLevel:
+			row[j] = string(v)
+		case DisplacementStatus:
+			row[j] = string(v)
+		case Gender:
+			row[j] = string(v)
 		default:
-			row[j] = value.(string)
+			row[j] = v.(string)
```
ludydoo, line 372: here, replace with fmt.Sprintf("%v", v). It will not panic if the value is not a string
```suggestion
			row[j] = fmt.Sprintf("%v", v)
```

ludydoo (approved):

PR #104, opened 2022-11-10
COR-588: files menu

internetti:
move file actions to navbar

### DIFF
--- a/web/templates/individuals.gohtml
+++ b/web/templates/individuals.gohtml
@@ -17,6 +17,8 @@
         const selectedRowAttributes = {"aria-selected": "true"}
         const selectedRowClasses = ["table-active"]
         const dataIndividualIdAttribute = "data-individual-id"
+        const activeDownloadButtonClasses = ["btn-outline-secondary"]
+        const inactiveDownloadButtonClasses = ["btn-outline-secondary", "disabled"]
         const activeDeleteButtonClasses = ["btn-danger"]
         const inactiveDeleteButtonClasses = ["btn-outline-secondary", "disabled"]
         const inactiveDeleteButtonText = "Delete individuals"
@@ -68,7 +70,7 @@
             if (count === 0) {
                 return inactiveDownloadButtonText
             } else {
-                return "Download selected individuals"
+                return "Download " + count + " individual" + (count > 1 ? "s" : "")
             }
         }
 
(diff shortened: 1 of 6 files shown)


web/templates/nav.gohtml
```diff
@@ -11,12 +26,45 @@
             <div id="navbarContent">
                 <div class="me-auto mb-2 mb-lg-0 w-100">
                     <div class="d-flex flex-row w-100 align-items-center">
-                        <a href="/static/nrc_grf_template.xlsx" download="nrc-grf-template" class="nav-link">
-                            Download template
-                        </a>
 
                         <div class="btn-group ms-5">
-                            <button type="button" class="btn btn-link text-secondary dropdown-toggle" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                            <button type="button" class="btn btn-link text-secondary dropdown-toggle px-2" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
ludydoo, line 31: This button is only valid for the individuals page. I think that we should rather add some javascript in the `individuals.gohtml` to dynamically create this element and insert it the navbar

neb42:
Why are we moving this to the nav bar? The ux seems wrong here

web/templates/individuals.gohtml
```diff
@@ -297,30 +316,21 @@
...
                             <span id="deleteIndividualsButtonContent">Delete individuals</span>
                         </button>
+
+                        <button onclick="downloadFunc()"
+                                style="width: 22rem"
+                                class="btn disabled btn-outline-secondary me-2 individual-action-button"
+                                id="downloadIndividualsButton">
+                            <i class="bi bi-download"></i>
+                            <span id="downloadIndividualsButtonContent">Download individuals</span>
+                        </button>
```
neb42, line 326: Should this still be here if it's being moved to the navbar?

web/templates/nav.gohtml
```diff
@@ -11,26 +26,63 @@
...
+                                Files
+                            </button>
+                            <div class="dropdown-menu">
+                                 <a href="/static/nrc_grf_template.xlsx"
+                                    download="nrc-grf-template"
+                                    class="dropdown-item"
+                                 >
+                                    Download template
+                                </a>
+                                {{if .RequestContext.SelectedCountry}}
```
neb42, line 56: This check needs to be more than "SelectedCountry". We need to check we are on the correct page

web/templates/individuals.gohtml
internetti: this one has a different purpose. the navbar one downloads the filtered (if no filters are selected, all) individuals, this one downloads the manually selected individuals.
I agree that this is not optimal, but I do think that the button to download the selected individuals should not be in the navbar.

web/templates/nav.gohtml
internetti: but the correct permissions are checked in the next line. and we need to show something if no country was selected yet. what do you mean by correct page?

neb42: The correct page is the individuals list. The download functionality doesn't make sense on any other page (which is why it is weird to have it in the navbar). The download template makes a bit more sense as it is constant.

PR #110, opened 2022-11-14
COR-652: add field for displacement status comment

internetti:


cmd/mockdata/main.go
```diff
@@ -231,6 +232,8 @@ func Generate(count uint) error {
 		}
 		displacementStatus := randomDisplacementStatus()
 
+		displacementStatusComment := randomText(f)
```
ludydoo, line 235: maybe check if `displacementStatus == "other"` to put a comment in there

ludydoo: otherwise good to go

neb42:
lgtm

PR #125, opened 2022-11-25
COR-668: show error message for invalid age

internetti:


pkg/views/forms/form.gohtml
```diff
@@ -72,11 +72,13 @@
     {{template "inputError" .}}
 {{end}}
 
+
 {{define "numberField"}}
     {{$field := .Field}}
     {{template "inputLabel" .}}
+    {{$field.Errors}}
```
neb42, line 79: Is this line to show the error? Why do we show the error here for number fields, but not for other field types?

internetti: ah no, that was for debugging ^^

web/templates/individual.gohtml
```diff
@@ -21,10 +21,10 @@
             const deleteButton = document.getElementById("deleteButton")
             const confirmDeleteButton = document.getElementById("confirmDeleteButton")
             submitButton.addEventListener("click", function () {
+                form.submit()
                 submitButton.disabled = true
                 deleteButton.disabled = true
                 confirmDeleteButton.disabled = true
-                form.submit()
```
neb42, line 27: Was this done in another pr?

neb42:
lgtm

web/templates/individual.gohtml
internetti: yes, had to do it here to to test, will fix by rebasing

PR #129, opened 2022-11-29
COR-676/COR-605: display errors in a nicer way

internetti:
this is improved by PR #173 / [COR-675]

[COR-675]: https://norwegianrefugeecouncil.atlassian.net/browse/COR-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


generated_error.csv
```diff
@@ -0,0 +1,17 @@
...ection_office,collection_agent_name,collection_agent_title,collection_time,comments,communication_disability_level,community_id,displacement_status,displacement_status_comment,email_1,email_2,email_3,full_name,free_field_1,free_field_2,free_field_3,free_field_4,free_field_5,sex,has_cognitive_disability,has_communication_disability,has_consented_to_rgpd,has_consented_to_referral,has_hearing_disability,has_mobility_disability,has_selfcare_disability,has_vision_disability,hearing_disability_level,household_id,identification_type_1,identification_type_explanation_1,identification_number_1,identification_type_2,identification_type_explanation_2,identification_number_2,identification_type_3,identification_type_explanation_3,identification_number_3,engagement_context,internal_id,is_head_of_community,is_head_of_household,is_minor,mobility_disability_level,nationality_1,nationality_2,phone_number_1,phone_number_2,phone_number_3,preferred_contact_method,preferred_contact_method_comments,preferred_name,preferred_communication_language,prefers_to_remain_anonymous,presents_protection_concerns,selfcare_disability_level,spoken_language_1,spoken_language_2,spoken_language_3,vision_disability_level
```
neb42, line 1: Probably not needed because of the other pr for testing files

internal/handlers/individuals_upload.go
```diff
@@ -25,8 +26,16 @@ func UploadHandler(individualRepo db.IndividualRepo) http.Handler {
 		var (
 			ctx = r.Context()
 			l   = logging.NewLogger(ctx)
+			//error = ""
```
neb42, line 29: Lots of comments that should be removed

internal/handlers/individuals_upload.go
```diff
@@ -50,25 +59,30 @@ func UploadHandler(individualRepo db.IndividualRepo) http.Handler {
...
-				http.Error(w, "failed to parse csv: "+err.Error(), http.StatusBadRequest)
+				//http.Error(w, "failed to parse csv: "+err.Error(), http.StatusBadRequest)
+				//error = "Could not parse csv: " + err.Error()
+				renderError("Could not parse csv: " + err.Error())
 				return
 			}
 		} else if strings.HasSuffix(filename, ".xlsx") || strings.HasSuffix(filename, ".xls") {
 			if err = api.UnmarshalIndividualsExcel(formFile, &individuals, &fields); err != nil {
 				l.Error("failed to parse excel file", zap.Error(err))
 				http.Error(w, "failed to parse excel file: "+err.Error(), http.StatusBadRequest)
```
neb42, line 70: We should probably render all upload errors consistently. I think this one and some others should use the same ui

neb42 (commented):
I think the UI is good, but a little off. Currently you replace the table with an inline modal, which is a half+half of two different styles. I think we should either:

Render a full page error message, with no table and a link back to the table. This is similar to what you have now except bigger, with no "x", and less modal styling (like the borders).

or

Render the table with the existing data, but add a modal over the top.

internal/handlers/individuals_upload.go
internetti: of course, I thought I had already pushed a later stage of the branch, sorry

internal/handlers/individuals_upload.go
```diff
@@ -86,25 +93,27 @@ func UploadHandler(individualRepo db.IndividualRepo) http.Handler {
...
-			http.Error(w, "Internal server error", http.StatusInternalServerError)
+			renderError("Could not load list of individuals: " + err.Error())
 			return
 		}
 
 		invalidIndividualIds := validateIndividualsExistInCountry(individualIds, existingIndividuals, selectedCountryID)
 		if len(invalidIndividualIds) > 0 {
 			l.Warn("user trying to update individuals that don't exist or are in the wrong country", zap.Strings("individual_ids", invalidIndividualIds))
-			http.Error(w, fmt.Sprintf("individuals not found: %v", invalidIndividualIds), http.StatusNotFound)
+			renderError(fmt.Sprintf("Could update individuals %s, they do not exist in the database for the selected country.", zap.Strings("individual_ids", invalidIndividualIds)))
```
neb42, line 103: Could not

neb42 (approved):
I think this looks good, but maybe needs a little bit of styling. That can be done separately though (and maybe dependant on your other pr)

Also conflicts

PR #130, opened 2022-12-01
switch country authorization from groups to nrc organisation

internetti:


deploy/oidc-users.json
```diff
@@ -28,7 +30,8 @@
     "name": "No country user",
     "nickname": "ncu",
     "password": "123",
-    "groups": []
+    "groups": [],
+    "nrcOrganisation": ""
```
neb42, line 34: I assume this would still be populated, but maybe with a value that doesn't match anything?

deploy/oidc-users.json
```diff
@@ -19,7 +20,8 @@
     "password": "123",
     "groups": [
       "NRC_Core_Uganda"
```
neb42, line 22: We should probably remove this group to ensure it's reading form the profile

internal/api/country.go
```diff
@@ -1,10 +1,11 @@
...
 
 type Country struct {
-	ID       string `db:"id"`
-	Code     string `db:"code"`
-	Name     string `db:"name"`
-	JwtGroup string `db:"jwt_group"`
+	ID              string `db:"id"`
+	Code            string `db:"code"`
+	Name            string `db:"name"`
+	JwtGroup        string `db:"jwt_group"`
```
neb42, line 7: I think we want to remove this as groups don't related to countries now

internal/db/migrations/postgres/018_add_nrc_organisation_to_country.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE countries
+
+    -- add nrc_organisation column
+    ADD COLUMN nrc_organisation      VARCHAR(255)         NOT NULL DEFAULT '';
```
neb42, line 4: Add a drop column here too

internal/handlers/country.go
```diff
@@ -15,13 +15,14 @@ import (
...
-		formParamName      = "Name"
-		formParamCode      = "Code"
-		formParamJwtGroup  = "JwtGroup"
+		templateName             = "country.gohtml"
+		newId                    = "new"
+		pathParamCountryID       = "country_id"
+		viewParamCountry         = "Country"
+		formParamName            = "Name"
+		formParamCode            = "Code"
+		formParamJwtGroup        = "JwtGroup"
```
neb42, line 24: Remove (I'll stop commenting this, but there's a bunch of places this will be removed from)

web/templates/country.gohtml
```diff
@@ -67,6 +67,29 @@
...
+                                       value="{{.Country.NrcOrganisation}}">
+
+                                <div class="form-text"
+                                     id="JwtGroupHelp">
+                                        The NRC Organisation that this country is associated with.
+                                        Users that belong to this NRC Organisation will be able to access this country's
+                                        data.
+                                </div>
+                            </div>
+                            <!-- End of JWT Group -->
```
neb42, line 92: A few `jwt` references in this block

neb42:
If I'm a global admin I should still be able to see the countries dropdown with 1 country

neb42:
Maybe revert the changes to the dropdown to push this through and do a new pr for it

web/templates/nav.gohtml
```diff
@@ -99,6 +99,10 @@
                             </div>
                         </div>
 
+                        {{ $auth := .RequestContext.Auth }}
+                        {{$countries := $auth.GetCountryIDsWithReadWritePermissions}}
+                        {{$countryNumber := len $countries}}
+                        {{if not (eq $countryNumber 1)}}
```
neb42, line 105: `{{if gt $countryNumber 1}` because we should handle when countries are 0

neb42 (approved):

PR #131, opened 2022-12-01
frontdoor

ludydoo:
Adds frontdoor and DNS zone for core. 
Variables have been added to the azure DevOps pipeline.
Once this is deployed, we need to add a `NS` record under `app.nrc.no`  for the `core-dev.app.nrc.no` DNS zone.

### DIFF
--- a/deploy/terraform/azurerm_cdn_frontdoor_route.fd.tf
+++ b/deploy/terraform/azurerm_cdn_frontdoor_route.fd.tf
@@ -1,0 +1,19 @@
+resource "azurerm_cdn_frontdoor_route" "backend" {
+  provider                      = azurerm.runtime
+  name                          = "backend"
+  cdn_frontdoor_endpoint_id     = azurerm_cdn_frontdoor_endpoint.backend.id
+  cdn_frontdoor_origin_group_id = azurerm_cdn_frontdoor_origin_group.backend.id
+  cdn_frontdoor_origin_ids      = [azurerm_cdn_frontdoor_origin.backend.id]
+  // cdn_frontdoor_rule_set_ids    = [azurerm_cdn_frontdoor_rule_set.fd.id]
+  enabled = true
+
+  forwarding_protocol    = "HttpsOnly"
+  https_redirect_enabled = true
+  patterns_to_match      = ["/*"]
+  supported_protocols    = ["Http", "Https"]
+
+  cdn_frontdoor_custom_domain_ids = [
+    azurerm_cdn_frontdoor_custom_domain.backend.id,
+  ]
+  link_to_default_domain = false
+}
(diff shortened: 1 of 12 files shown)


deploy/terraform/azurerm_cdn_frontdoor_route.fd.tf
```diff
@@ -0,0 +1,19 @@
+resource "azurerm_cdn_frontdoor_route" "backend" {
+  provider                      = azurerm.runtime
+  name                          = "backend"
+  cdn_frontdoor_endpoint_id     = azurerm_cdn_frontdoor_endpoint.backend.id
+  cdn_frontdoor_origin_group_id = azurerm_cdn_frontdoor_origin_group.backend.id
+  cdn_frontdoor_origin_ids      = [azurerm_cdn_frontdoor_origin.backend.id]
+  // cdn_frontdoor_rule_set_ids    = [azurerm_cdn_frontdoor_rule_set.fd.id]
```
neb42, line 7: Is this for the ip restriction?

neb42:
For the url can we do `uat.core.app.nrc.no` so that when people go to prod and type `co` it doesn't autocomplete uat first

neb42:
We talked about not having frontdoor on dev. I can't see anything here to configure that.

deploy/terraform/azurerm_cdn_frontdoor_route.fd.tf
ludydoo: No, its a placeholder if we ever want to put frontdoor rules.

ludydoo:
Its using frontdoor standard with is 40$ a month instead of 330

neb42:
> Its using frontdoor standard with is 40$ a month instead of 330

At a glance premium doesn't seem necessary for uat/prod. Do you agree?

ludydoo:
> For the url can we do `uat.core.app.nrc.no` so that when people go to prod and type `co` it doesn't autocomplete uat first

This is gonna get messy. I'd rather do dev-core.app.nrc.no. otherwise you'll get a weird subdomain hierarchy going on

neb42 (approved):

ludydoo:
> > Its using frontdoor standard with is 40$ a month instead of 330
> 
> At a glance premium doesn't seem necessary for uat/prod. Do you agree?

For prod, premium is needed. There's a powerful WAF included.

deploy/terraform/azurerm_front_door_profile.fd.tf
```diff
@@ -0,0 +1,6 @@
+resource "azurerm_cdn_frontdoor_profile" "fd" {
+  provider            = azurerm.runtime
+  name                = "fd-${var.app_name}-${var.environment}"
+  resource_group_name = azurerm_resource_group.rg.name
+  sku_name            = "Standard_AzureFrontDoor"
```
neb42, line 5: This should be a variable if we want premium for prod

PR #137, opened 2022-12-01
fix migrations

internetti:


internal/db/migrations/postgres/018_add_nrc_organisation_to_country.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE countries
+    -- drop jwt_group column
+    DROP COLUMN jwt_group;
```
neb42, line 3: `IF EXISTS`

internal/db/migrations/postgres/018_add_nrc_organisation_to_country.up.sql
```diff
@@ -0,0 +1,6 @@
...
+ALTER TABLE countries
+    -- add nrc_organisation column
+    ADD COLUMN nrc_organisation      VARCHAR(255)         NOT NULL DEFAULT '';
```
neb42, line 6: `IF NOT EXISTS`

internal/db/migrations/postgres/018_add_nrc_organisation_to_country.up.sql
ludydoo: > `IF EXISTS`

I don't think this command exists for columns

internal/db/migrations/postgres/018_add_nrc_organisation_to_country.up.sql
ludydoo: Same comment as above

ludydoo: Maybe I'm wrong

ludydoo: I'm wrong.

ludydoo (approved):

PR #139, opened 2022-12-01
update frontdoor cache config

ludydoo:
As per the docs

https://learn.microsoft.com/en-us/azure/app-service/overview-authentication-authorization
https://learn.microsoft.com/en-us/azure/static-web-apps/front-door-manual#disable-cache-for-auth-workflow


neb42 (approved):
Looks good, but should other routes be added to the cache rule?

ludydoo:
> Looks good, but should other routes be added to the cache rule?

Not that I can think of right now. If the need arises to configure the cache in a different way, this is an example on how to do so

PR #146, opened 2022-12-02
materialize individual file download

ludydoo:


neb42:
The download also contains query parameter. Should the filename be a hash of the country id, extension, and query to ensure uniqueness? 

Happy to merge this now to test and add that in later

neb42:
oh nvm, it's generating a new id per download

neb42 (approved):

PR #150, opened 2022-12-05
add individual indices

ludydoo:
postgres will most likely use only 1 index at a time.
Added 2 indices
1. `country_id` and `deleted_at`. This is the arguments used for the default search terms
2. another index with intuitively most used search fields. Used to speed up search queries.

### DIFF
--- a/internal/db/migrations/postgres/019_individual_indices.up.sql
+++ b/internal/db/migrations/postgres/019_individual_indices.up.sql
@@ -1,21 +1,35 @@
 -- Add indices on the individual_registrations table
-
 CREATE INDEX IF NOT EXISTS idx_individual_registrations__address ON individual_registrations (address);
 CREATE INDEX IF NOT EXISTS idx_individual_registrations__age ON individual_registrations (age);
 CREATE INDEX IF NOT EXISTS idx_individual_registrations__birth_date ON individual_registrations (birth_date);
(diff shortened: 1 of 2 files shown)


internal/db/migrations/postgres/019_individual_indices.up.sql
```diff
@@ -0,0 +1,35 @@
+-- Add indices on the individual_registrations table
+
```
neb42, line 2: `created_at` and `updated_at`, seen as they can be sorted on? Would an index on `deleted_at` help seen as we filter every query on it?

ludydoo: sure, also adding other fields in the search box

ludydoo: actually we should include the country_id in all indices, since only 1 index can be used at a time

internal/db/migrations/postgres/019_individual_indices.up.sql
```diff
@@ -0,0 +1,34 @@
...
+         phone_number_1,
+         phone_number_2,
+         phone_number_3,
+         identification_number_1,
+         identification_number_2,
+         identification_number_3,
+         is_minor,
+         household_id,
+         community_id
+            );
```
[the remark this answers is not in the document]
ludydoo: also this index would be **massive** with millions of records

neb42 (approved):
These changes will only effect the individuals list, but I think that's ok as that's where the main issues are.

PR #152, opened 2022-12-05
allow multiple nrc organisations

internetti:


web/templates/country.gohtml
```diff
@@ -45,24 +37,36 @@
...
+                                       aria-labelledby="NrcOrganisationsLabel"
+                                       value="{{.Country.NrcOrganisations}}">
                                 <div class="form-text"
-                                     id="NrcOrganisationHelp">
+                                     id="NrcOrganisationsHelp">
                                         The NRC Organisation that this country is associated with.
                                         Users that belong to this NRC Organisation will be able to access this country's
                                         data.
+                                    <br>
+                                    If you want to enter multiple NRC Organisations, please separate them with a comma, <b>without adding extra spaces</b>.
```
neb42, line 67: We should trim the strings

neb42: We should remove the comment about extra spaces

ludydoo: Another possibility would be to add a button in javascript that would append new fields for the NrcOrganisation. If a form containing multiple fields with the same name are POSTed, the form data will contain 
```
var nrcOrganisations []string := r.Form["nrcOrganisation"]
```

Then the user would not need to comma-separate, it would be handled by the built-in mechanisms of html forms

ludydoo: eg. 
```
<input name="nrcOrganisation" value="nrcOrg1"/>
<input name="nrcOrganisation" value="nrcOrg2"/>
etc.
```

internal/handlers/country.go
```diff
@@ -59,7 +60,13 @@ func HandleCountry(templates map[string]*template.Template, repo db.CountryRepo)
 		}
 		country.Name = strings.TrimSpace(r.FormValue(formParamName))
 		country.Code = strings.TrimSpace(strings.ToLower(r.FormValue(formParamCode)))
-		country.NrcOrganisation = strings.TrimSpace(r.FormValue(formParamNrcOrganisation))
+		orgs := containers.NewStringSet(r.Form[formParamNrcOrganisation]...)
```
neb42, line 63: Do we need to trim these items?

neb42:
Small comment but otherwise looks good

neb42 (approved):

PR #154, opened 2022-12-12
introduce cache to cache language options

internetti:
add a cache to read language options from.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/nrc-no/notcore
 go 1.18
 
 require (
+	github.com/allegro/bigcache/v3 v3.1.0
 	github.com/coreos/go-oidc/v3 v3.4.0
 	github.com/golang/mock v1.6.0
 	github.com/google/uuid v1.3.0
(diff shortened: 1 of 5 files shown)


ludydoo:
I don't think a cache is needed for this, since this is a read-only thing and is known at compile time. This could be saved within a var, and initialized in an `init` block. Example:

```
// init runs when the program start
var languageOptions []forms.SelectInputFieldOption
func init(){
  for _, l := range languages.Languages {
    languageOptions = append(languageOptions, Option{...})
  }
}

func buildLanguageOptions() []forms.SelectInputFieldOption{
  return languageOptions
}

```

ludydoo:
But we should really filter out the languages that we allow, instead of letting the user select among ~20k languages. In `internal/constants/languages.go`, perhaps having a list of 

`enabled_languages.json`
```
{"fr":{}, "en":{},...}
```

`languages.go`
```
var allLanguages Language[]
var enabledLanguages
func init(){
   if err := json.Unmarshal([]byte(enabledLanguagesJson), &enabledLanguages); err != nil {
		panic(err)
	}
   if err := json.Unmarshal([]byte(languagesJson), &allLanguages); err != nil {
		panic(err)
	}
	for _, l := range allLanguages{
	   if _, ok := enabledLanguages[l.Name]; ok {
	     Languages = append(Languages, l)
	   }
	}
	
}
```

PR #156, opened 2022-12-13
create a waf policy for the backend

ludydoo:


deploy/terraform/azurerm_web_application_firewall_policy.backend.tf
```diff
@@ -0,0 +1,17 @@
+resource "azurerm_web_application_firewall_policy" "backend" {
+  name                = "waf${random_id.app_id.hex}"
+  resource_group_name = azurerm_resource_group.rg.name
+  location            = azurerm_resource_group.rg.location
+
+  policy_settings {
+    file_upload_limit_in_mb     = 500
```
internetti, line 7: are we sure this is enough?

PR #159, opened 2022-12-23
Fetch groups from user info endpoint

neb42:
Was having some issues getting groups from the id token on my okta dev server. Seen as we already use the user info endpoint, then this seems a more stable place to fetch this info. 

I also seem to recall there also being an okta bug about not fetching groups with the refresh token. This would fix that.


neb42:
Going to merge this as everyone is away and I want to deploy to test it

PR #162, opened 2023-01-10
COR-622: (de)activate action in table and search

internetti:
add new column in individual_registrations, for the activate boolean, true by default
also added as a filter to the search
the actions delete, activate and deactivate have been moved to a meatball menu next to the all-select-checkbox in the table.
inactive individuals will only be shown if explicitly requested

includes [COR-624], [COR-625] and [COR-621]

[COR-624]: https://norwegianrefugeecouncil.atlassian.net/browse/COR-624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[COR-625]: https://norwegianrefugeecouncil.atlassian.net/browse/COR-625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[COR-621]: https://norwegianrefugeecouncil.atlassian.net/browse/COR-621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/templates/searchForm.gohtml
```diff
@@ -74,6 +74,40 @@
...
+                               id="ActiveNo"
+                               name="active"
+                               {{if .Options.IsNotActiveSelected }}checked="checked"{{end}}>
+                        <label class="form-check-label" for="ActiveNo">
+                            No
+                        </label>
+                    </div>
+                </div>
+            </div>
+            <!-- End Is Active -->
```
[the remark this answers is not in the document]
neb42: And show only active by default

web/templates/individuals.gohtml
```diff
@@ -338,7 +351,44 @@
...
+                                    <li class="list-group-item"
+                                        onclick="submitActionForm('activate')"
+                                        data-toggle="tooltip" data-placement="top" title="Activate selected individuals"
+                                        id="activateButton"
+                                    >
+                                        <i class="bi bi-eye me-2"></i>
+                                        <span id="activateIndividualsButtonContent">Activate individuals</span>
+                                    </li>
+                                </ul>
+                            </div>
```
neb42, line 388: I think this looks cool, but is super hidden. Maybe keep the old delete button, but use the same component as the download button?

web/templates/individuals.gohtml
```diff
@@ -614,8 +664,15 @@
                             class="individual-row"
                             data-individual-id="{{.ID}}">
 
+                            <td class="sticky-column text-center" style="width: 1rem" >
+                                {{if not .Active}}
+                                    <div class="">
+                                        <i class="bi bi-eye-slash-fill "></i>
+                                    </div>
+                                {{end}}
+                            </td>
```
neb42, line 673: I like this, but maybe worth checking with Fix if it's clear enough

internal/db/individual.go
```diff
@@ -22,8 +22,14 @@ type IndividualRepo interface {
...
 	PutMany(ctx context.Context, individuals []*api.Individual, fields containers.StringSet) ([]*api.Individual, error)
+	PerformAction(ctx context.Context, id string, action string) error
+	PerformActionMany(ctx context.Context, ids containers.StringSet, action string) error
 	SoftDelete(ctx context.Context, id string) error
 	SoftDeleteMany(ctx context.Context, ids containers.StringSet) error
+	Activate(ctx context.Context, id string) error
+	ActivateMany(ctx context.Context, ids containers.StringSet) error
+	Deactivate(ctx context.Context, id string) error
+	DeactivateMany(ctx context.Context, ids containers.StringSet) error
 }
```
neb42, line 33: Actually it's these that aren't used now

internal/db/const.go
```diff
@@ -1,4 +1,40 @@
...
+	targetField: "deleted_at",
+	newValue:    time.Now().UTC().Format(time.RFC3339),
+}
+var activateAction = individualAction{
+	conditions:  []string{"and active IS false"},
+	targetField: "active",
+	newValue:    true,
+}
+var deactivateAction = individualAction{
+	conditions:  []string{"and active IS true"},
```
neb42, line 25: The conditions here cause an error:

* In the table select 2 rows, one is inactive and the other is active
* Try to activate or deactivate both
* The filter causes the `rowsAffected` to be `1` not `2`
* There is then a safety check that `rowsAffected = len(ids)`, this fails and errors

I think we just remove the conditions here, but keep the delete condition

neb42 (changes requested):

internal/api/individual_tabular.go
```diff
@@ -94,6 +94,8 @@ func (i *Individual) unmarshalTabularData(colMapping map[string]int, cols []stri
 		switch field {
 		case constants.FileColumnIndividualID:
 			i.ID = cols[idx]
+		case constants.FileColumnIndividualActive:
+			i.Active = isTrue(cols[idx])
```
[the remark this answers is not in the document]
internetti: there's a ticket to enable downloading inactive individuals, so the possibility should be there, but yeah, not by default

internal/api/individual_tabular.go
```diff
@@ -94,6 +94,8 @@ func (i *Individual) unmarshalTabularData(colMapping map[string]int, cols []stri
 		switch field {
 		case constants.FileColumnIndividualID:
 			i.ID = cols[idx]
+		case constants.FileColumnIndividualInactive:
+			i.Inactive = isFalse(cols[idx])
```
neb42, line 98: Why are we looking for false values rather than true?

internal/db/individual.go
```diff
@@ -305,6 +308,14 @@ func (i individualRepo) softDeleteManyInternal(ctx context.Context, tx *sqlx.Tx,
...
 
+func (i individualRepo) SoftDelete(ctx context.Context, id string) error {
+	return i.PerformAction(ctx, id, DeleteAction)
+}
+
+func (i individualRepo) SoftDeleteMany(ctx context.Context, ids containers.StringSet) error {
+	return i.PerformActionMany(ctx, ids, DeleteAction)
+}
+
```
neb42, line 318: These should be deleted too

internal/db/migrations/postgres/022_add_inactive_column_to_individual.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE individual_registrations
+    DROP COLUMN IF EXISTS active;
```
neb42, line 2: I don't think this is needed. Just teardown your local db

internal/db/sql_individual_list.go
```diff
@@ -87,6 +88,15 @@ func newGetAllIndividualsSQLQuery(driverName string, options api.ListIndividuals
 	return qry
 }
 
+func (g *getAllIndividualsSQLQuery) withInactive(inactive *bool) *getAllIndividualsSQLQuery {
+	if inactive == nil || *inactive != true {
+		g.writeString(" AND inactive = false")
+		//} else {
+		//	g.writeString(" AND inactive = ").writeArg(true)
```
neb42, line 95: Delete

internal/api/individual_list_options.go
```diff
@@ -141,6 +142,14 @@ func (s *SortTerms) parseTerm(term string) (string, SortDirection, error) {
 	return term, direction, nil
 }
 
+func (o ListIndividualsOptions) IsInactiveSelected() bool {
+	return *o.Inactive == true
+}
+
+func (o ListIndividualsOptions) IsNotInactiveSelected() bool {
+	return o.Inactive != nil && !*o.Inactive
```
neb42, line 150: Should this be an `or`?

neb42 (changes requested):
* Delete button styling is broken
* I think all the buttons need to have text for clarity. Currently we don't say active/inactive/deactivate anywhere other than tooltips
* I'd move the inactive table filter to the search panel
* There's still an error when performing a batch action on a mix of active + inactive. I assume because there's a filter for `inactive = false` being applied somewhere. Actually this also happens when trying to activate a single individual within the list view (again probably because of the filter).

internal/api/individual_tabular.go
internetti: because we have 3 possible values, true, false and nil. true and nil both mean not including inactive users. only false means including them. so it made more sense to explicitly look for false values.

cmd/template.go
```diff
@@ -28,6 +28,7 @@ to quickly create a Cobra application.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		// Commented out fields are currently blank, but left in place in case we want to change values later
 		individual := &api.Individual{
+			Inactive:                      false,
```
neb42, line 31: I'd remove inactive from the template. I think when creating a new file they won't use the inactive field

neb42 (commented):
* Delete button on the individual form still has broken css
* Activating/deactivating in the table doesn't preserve the search query

neb42:
conflicts

internetti:
@neb42 can you give me a screenshot of the broken css of the delete button? I don't see anything wrong

internetti:
@neb42 about the delete button styling, you might have to run another `make bootstrap`

neb42:
The css was down to `make bootstrap`

The query string still isn't being preserved. I think the change is needed in internal/handlers/individuals_action.go, with this block:

```
if options.Inactive != nil && *options.Inactive {
  http.Redirect(w, r, fmt.Sprintf("/countries/%s/individuals?inactive=true", countryID), http.StatusFound)
} else {
  http.Redirect(w, r, fmt.Sprintf("/countries/%s/individuals", countryID), http.StatusFound)
}
```

neb42 (approved):

PR #166, opened 2023-01-17
COR-626: differentiate read, write and admin permissions

internetti:
added permission groups for read and write permissions.
the permission levels are hierarchical, i.e. `global admin` > `write` > `read`
the levels are then used to disable/enable features in the frontend

I have added the variables `jwt_can_read_group` and `jwt_can_write_group` in azure, but I could only set the scope to `Release`, that will have to be adjusted


web/templates/individuals.gohtml
```diff
@@ -334,12 +334,16 @@
         </div>
 
         <!-- INDIVIDUALS TABLE -->
-        <form id="individuals-action-form" class="table-responsive opacity-0 scroll-body" method="post">
+        <form id="individuals-action-form" class="table-responsive scroll-body" method="post">
```
neb42, line 337: Why remove `opacity-0`?

web/templates/individuals.gohtml
```diff
@@ -334,12 +334,16 @@
...
             <table class="table table-individuals mb-0">
                 <thead>
                 <tr>
                     <th class="sticky-column text-center" style="width: {{$selectWidth}}">
-                        <input class="form-check-input" type="checkbox" id="selectAllCheckbox"/>
+                        <input class="form-check-input"
+                               type="checkbox"
+                               id="selectAllCheckbox"
+                               {{if not $canWrite}}disabled{{end}}
+                        />
```
neb42, line 346: Probably better to remove this column if they can't write

internal/auth/interface.go
```diff
@@ -16,56 +16,76 @@ const (
 
 type Interface interface {
 	IsGlobalAdmin() bool
-	GetCountryIDsWithReadWritePermissions() containers.StringSet
+	GetAllowedCountries() containers.StringSet
 	HasCountryLevelPermission(countryID string, perm Permission) bool
```
neb42, line 20: Maybe replace uses of this with the new methods

web/templates/nav.gohtml
```diff
@@ -52,60 +52,58 @@
...
-                                              method="post"
-                                        >
-                                            <input
-                                                    hidden
-                                                    id="individual-upload--file"
-                                                    name="file"
-                                                    type="file"
-                                                    class="form-control"
-                                                    accept="text/csv, application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+                            {{if (.RequestContext.Auth.HasMinimumPermissionWrite .RequestContext.SelectedCountry.ID)}}
```
neb42, line 67: Maybe a request context method like `.RequestContext.HasSelectedCountryWritePermission` could clean these uses up a bit

internal/utils/types.go
```diff
@@ -0,0 +1,17 @@
+package utils
```
neb42, line 1: Is there a better place these structs could live? I generally don't like putting things in a top level `utils` and also `types.go` is super generic.

neb42 (changes requested):

internal/utils/types.go
internetti: the permissions type is only used in one file, I'll move it there. the other one is used in all kinds of places, I don't know where better to put it

web/templates/individuals.gohtml
internetti: this class is removed immediately anyway, it was some kind of hack by ludo to prevent the jumping of the table, but I couldn't observe any jumping anymore, so this hack is obsolete now.

internal/server/middleware/permissions.go
```diff
@@ -69,15 +63,27 @@ func parsePermissions(allCountries []*api.Country, globalAdminGroup string, user
...
 			continue
 		}
+		if group == jwtGroups.CanRead {
+			canRead = true
+			continue
+		}
+		if group == jwtGroups.CanWrite {
+			canWrite = true
+			continue
+		}
```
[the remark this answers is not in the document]
internetti: the hierarchy was handled in the auth interface, but moving it here does make it a bit cleaner

internal/handlers/templates.go
```diff
@@ -105,6 +105,22 @@ type RequestContext struct {
 	Session auth.Session
 }
 
+func (r RequestContext) HasSelectedCountryWritePermission() bool {
+	authInterface, err := utils.GetAuthContext(r.Request.Context())
```
neb42, line 109: I don't think this is needed. Can you do `r.Auth.HasCountryPermissionWrite` etc

web/templates/individuals.gohtml
```diff
@@ -849,18 +861,21 @@
...
-                    data-bs-toggle="modal"
-                    data-bs-target="#deleteIndividualsConfirmationModal"
-                    data-toggle="tooltip" data-placement="top" title="Delete selected individuals"
-                    onclick="refreshDeleteConfirmationModal()">
-                <i class="bi bi-trash"></i>
-                <span id="deleteIndividualsButtonContent">Delete individuals</span>
-            </button>
-        {{end}}
+        <div>
+            {{ if or .RequestContext.Auth.IsGlobalAdmin .RequestContext.HasSelectedCountryWritePermission}}
```
neb42, line 865: Maybe `HasSelectedCountryWritePermission` should return true if you're a global admin? Because that is the case

neb42: The global admin check doesn't seem to be there for any other places where you need write permissions

neb42 (commented):
Before merging this it's probably worth setting up okta groups and azure dev ops variables

internal/handlers/templates.go
internetti: I tried that first, threw an error for some reason. now it works. weird.

internetti:
@neb42 the variables are already set up, at least for the development stage. about the okta groups, how free are we with naming them? I guess for development we can do what we want, but for the others, do we need to consult Mads?

neb42:
Tests are failing (I just triggered the build manually)

neb42 (approved):

PR #169, opened 2023-01-18
COR-683: Add first/middle/last name fields

neb42:


internetti (commented):
what do you think about the name search input searching in all name fields?

internetti (approved):

PR #170, opened 2023-01-19
COR-681: iso values

internetti:
looking up country names and language names when downloading and allowing iso codes as well as names for countries and languages when uploading


neb42 (commented):
If I upload an invalid value it is treated as an empty string rather than an error

neb42:
I think this will now treat empty strings as an error. I think it should only error if the column is populated and invalid.

neb42 (approved):

PR #172, opened 2023-01-23
COR-680: explicit list button & cancel button

internetti:
added a cancel button on the individual form and a more explicit link back to the individuals list
took the opportunity to visually slim the navigation a bit, let me know what you think

### DIFF
--- a/web/templates/individual.gohtml
+++ b/web/templates/individual.gohtml
@@ -14,6 +14,10 @@
             return input.type !== "hidden" && !input.disabled
         }
 
+        function goBack () {
+            window.history.back();
+        }
+
         document.addEventListener("DOMContentLoaded", function () {
             const form = document.getElementById("individualForm")
             const deletionForm = document.getElementById("delete-individual-form")
@@ -96,6 +100,9 @@
                 <!-- END DELETE BUTTON -->
             {{end}}
             <div class="flex-grow-1"></div>
+            <!-- CANCEL BUTTON -->
+            <button onclick="goBack()" class="btn btn-outline-primary mx-4">Cancel</button>
+            <!-- END CANCEL BUTTON -->
             <!-- SAVE BUTTON -->
             <button id="submitButton" type="submit" class="btn btn-primary mx-4">Save</button>
             <!-- END SAVE BUTTON -->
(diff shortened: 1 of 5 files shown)


web/templates/nav.gohtml
```diff
@@ -38,36 +38,35 @@
...
-    <nav class="navbar p-5">
-        <div class="container-fluid" id="navbar-container">
+    <nav class="navbar py-3 px-5">
+        <div class="container-fluid">
             <a class="home-link me-0 me-lg-2" href="{{if eq "" .RequestContext.SelectedCountryID}}/{{else}}/countries/{{.RequestContext.SelectedCountryID}}/individuals{{end}}">
                 <div class="d-flex flex-row align-items-center me-5">
-                    <img src="/static/nrc_logo.png" alt="nrc logo" width="65" height="65" class="d-inline-block"/>
-                    <span class="ms-3">Core{{if .RequestContext.SelectedCountry}} {{.RequestContext.SelectedCountry.Name}}{{end}}</span>
+                    <img src="/static/nrc_logo.png" alt="nrc logo" width="50" height="50" class="d-inline-block"/>
+                    <span class="ms-3">{{if .RequestContext.SelectedCountry}}{{.RequestContext.SelectedCountry.Name}} Participants{{else}}Core{{end}}</span>
```
neb42, line 46: I don't think this is sufficient. I think part of the issue is that this doesn't look like a navigation button and just looks like a logo. I think it's also important to keep the "core" title. We probably need a new link on the right hand side

web/templates/individual.gohtml
```diff
@@ -93,6 +97,9 @@
                 <!-- END DELETE BUTTON -->
             {{end}}
             <div class="flex-grow-1"></div>
+            <!-- CANCEL BUTTON -->
+            <button onclick="goBack()" class="btn btn-outline-primary mx-4">Cancel</button>
+            <!-- END CANCEL BUTTON -->
```
neb42, line 102: I'd say this is maybe redundant if we have the link in my other comment and it would reduce clutter. Not a strong opinion though

neb42 (commented):
conflicts

web/templates/individual.gohtml
internetti: I wouldn't insist on it either, but it's part of the ticket

neb42:
Any reason for the font size change in the navigation?

internetti:
@neb42 with one more item in the navbar, I felt things were getting too crowded

neb42 (approved):

PR #173, opened 2023-01-24
COR-675: displaying all errors at once

internetti:
show multiple errors, grouped by category
move error display to dedicated template


neb42 (approved):
Possibly my favourite pr so far. My own comment would be that the error message indicates an error in the excel row including the header row. I felt a little confused by this, but I can also see how the other way around would be confusing.

I also found a bunch of (small) upload issues when testing this, but I'll open a different pr.

neb42:
tests are failing

PR #176, opened 2023-01-24
Add service fields

neb42:
Tests are failing upstream so I'll fix after merging

### DIFF
--- a/internal/api/service_cc.go
+++ b/internal/api/service_cc.go
@@ -1,0 +1,121 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/nrc-no/notcore/internal/containers"
+	"github.com/nrc-no/notcore/pkg/logutils"
+)
+
+type ServiceCC string
+
+const (
+	ServiceCCNone       ServiceCC = ""
+	ServiceCCShelter    ServiceCC = "shelter_and_settlements"
+	ServiceCCWash       ServiceCC = "wash"
+	ServiceCCProtection ServiceCC = "protection"
+	ServiceCCEducation  ServiceCC = "education"
+	ServiceCCICLA       ServiceCC = "icla"
+	ServiceCCLFS        ServiceCC = "lfs"
+	ServiceCCCVA        ServiceCC = "cva"
+	ServiceCCOther      ServiceCC = "other"
+)
+
+func AllServiceCCs() containers.Set[ServiceCC] {
+	return containers.NewSet[ServiceCC](
+		ServiceCCNone,
+		ServiceCCShelter,
+		ServiceCCWash,
+		ServiceCCProtection,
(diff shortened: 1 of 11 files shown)


internal/api/service_cc.go
```diff
@@ -0,0 +1,121 @@
...
+	case string(ServiceCCICLA):
+		return ServiceCCICLA, nil
+	case string(ServiceCCLFS):
+		return ServiceCCLFS, nil
+	case string(ServiceCCCVA):
+		return ServiceCCCVA, nil
+	case string(ServiceCCOther):
+		return ServiceCCOther, nil
+	default:
+		return "", fmt.Errorf("unknown displacement status type: %v", logutils.Escape(str))
```
internetti, line 89:
```suggestion
		return "", fmt.Errorf("unknown service type: %v", logutils.Escape(str))
```

internal/api/validation/individual.go
```diff
@@ -82,6 +82,20 @@ func validateIndividual(i *api.Individual, p *validation.Path) validation.ErrorL
 	allErrs = append(allErrs, validateIndividualSpokenLanguage(i.SpokenLanguage1, p.Child("spokenLanguage1"))...)
 	allErrs = append(allErrs, validateIndividualSpokenLanguage(i.SpokenLanguage2, p.Child("spokenLanguage2"))...)
 	allErrs = append(allErrs, validateIndividualSpokenLanguage(i.SpokenLanguage3, p.Child("spokenLanguage3"))...)
+	allErrs = append(allErrs, validateIndividualServiceCC(i.ServiceCC1, p.Child("serviceCC1"))...)
```
internetti, line 85: should we not validate the request and delivery dates? birthday and collection time are validated

internetti (changes requested):

internal/api/validation/individual.go
```diff
@@ -412,6 +426,20 @@ func validateIndividualServiceCC(cc api.ServiceCC, path *validation.Path) valida
 	return validation.ErrorList{validation.NotSupported(path, cc, allowedServiceCCsStr)}
 }
 
+func validateIndividualServiceDate(d *time.Time, path *validation.Path) validation.ErrorList {
+	allErrs := validation.ErrorList{}
+	if d == nil {
+		return allErrs
+	}
+	if d.After(time.Now()) {
+		allErrs = append(allErrs, validation.Invalid(path, d, "birthdate cannot be in the future"))
```
internetti, line 435: the log message still says birthday.
also, it might be useful to deduplicate all the date validation functions, but that could be done later.

neb42: Yeah I did think that, but it could be that different dates have different requirements. Eg. some are required, some can be in the future (like a deadline)

internetti (approved):

PR #182, opened 2023-02-06
COR-621: rework inactive query logic

internetti:
it was requested that the inactive field should have the options of yes, no, and all, like the other boolean search fields.
and it was supposed to be sortable in the table.


neb42 (commented):
conflict

neb42 (approved):

PR #184, opened 2023-02-08
COR-724/725: add pwd fields

internetti:


internal/db/migrations.go
```diff
@@ -46,6 +46,7 @@ var migrations = []migration{
 	migrationFromFile("022_new_individual_fields"),
 	migrationFromFile("023_add_service_fields"),
 	migrationFromFile("024_add_inactive_column_to_individual"),
+	migrationFromFile("025_add_general_disability_fields"),
```
neb42, line 49: Be careful when merging this and other prs

web/templates/individuals.gohtml
```diff
@@ -761,6 +769,19 @@
                             </td>
                             <!-- End Email -->
 
+                            <!-- HasDisability -->
```
neb42, line 772: Do we want this in the table? None of the other disability columns are in the table.

internetti: according to the ticket and the huge excel file defining the criteria, yes

neb42 (approved):

PR #185, opened 2023-02-09
COR-676: form/upload validation

internetti:
some of the select fields were still accepting random values on upload
booleans were not checked for valid values on upload
birthdays were not checked on minimum valid value on upload

added the missing validations and unified the rest


internal/api/enumTypes/country.go
```diff
@@ -1,4 +1,4 @@
-package api
+package enumTypes
```
neb42, line 1: Why has country moved?

internal/api/enumTypes/disability_level.go
```diff
@@ -49,14 +49,24 @@ func ParseDisabilityLevel(str string) (DisabilityLevel, error) {
 	switch str {
 	case string(DisabilityLevelNone):
 		return DisabilityLevelNone, nil
+	case DisabilityLevelNone.String():
+		return DisabilityLevelNone, nil
```
neb42, line 53:
```
case string(DisabilityLevelNone), DisabilityLevelNone.String():
  return DisabilityLevelNone, nil
```

internal/api/enumTypes/country.go
internetti: I classified it as a list of options for the form, but you're right, doesn't belong here, I'll move it back

internetti:
@neb42 the enumTypes folder is not necessary per se, I just felt that these types do play a different role than the other ones in that folder, so I wanted to group them together to find them more easily

PR #186, opened 2023-02-10
COR-685/719/720/722: display fields

internetti:
add the fields for identification numbers, services with their dates, minor and female headed household to table and search form

the services have one common search field


neb42 (approved):
Happy to approve this for now but I think the table is becoming over populated

PR #201, opened 2023-02-15
COR-744: show de/activation buttons at appropriate times

internetti:


neb42 (commented):
Could we just disabled the buttons instead? It'd make the ui feel more consistent

neb42 (approved):

PR #202, opened 2023-02-16
COR-747: remove logout button

internetti:
we had talked about removing it in the past since it behaves weirdly and is unneccessary


neb42 (commented):
We should probably remove all uses of the logout url in the code

neb42 (approved):

PR #205, opened 2023-02-16
mark active nav link

internetti:
when we're on the participants page, mark that nav link as active


neb42 (commented):
I think the right way to do this is to add the `active` class and have css for `.nav-link .active`

neb42 (approved):

PR #208, opened 2023-02-24
COR-754: limit upload to 10.000 lines

internetti:


internal/api/individual_tabular.go
```diff
@@ -66,6 +68,11 @@ func UnmarshalIndividualsExcel(reader io.Reader, individuals *[]*Individual, fie
 }
 
 func UnmarshalIndividualsTabularData(data [][]string, individuals *[]*Individual, fields *[]string) []FileError {
+
+	if len(data[1:]) > UPLOAD_LIMIT {
```
neb42, line 72: I think if we do the check here it'd be better to pass in `UPLOAD_LIMIT` as a `row_limit` argument that can be a nil pointer. Then have the handler set the row limit.

internetti: @neb42 should we maybe even set that on the server level? depending on the server config, we could handle more uploads

neb42: I'd say no because I don't think scaling resources would affect this (unless we moved to a better performing cpu, not more cpus). But it could make it easier to tweak if it's something we think will need adjusting

PR #212, opened 2023-03-16
COR-755: monitoring

internetti:


azure-pipelines.yml
```diff
@@ -1,5 +1,5 @@
 trigger:
-- main
+- COR-755_monitoring
```
internetti, line 2: these will have to be changed back before merging

neb42:
copilot:all

internetti:
> copilot:all

is that a command? what does that do?

deploy/terraform/azurerm_monitor_smart_detector.tf
```diff
@@ -0,0 +1,107 @@
...
+    resource_id             = azurerm_logic_app_workflow.logic-app-teams.id
+    callback_url            = azurerm_logic_app_workflow.logic-app-teams.access_endpoint
+  }
+}
+
+resource "azurerm_monitor_smart_detector_alert_rule" "dep-latency" {
+  name                = "Dependency Latency Degradation"
+  provider            = azurerm.runtime
+  resource_group_name = azurerm_resource_group.rg.name
+  severity            = "Sev3"
```
neb42, line 29: What do the different severity levels mean?

deploy/terraform/azurerm_monitor_smart_detector.tf
```diff
@@ -0,0 +1,107 @@
...
+  scope_resource_ids  = [azurerm_application_insights.aisd.id]
+  frequency           = "P1D"
```
neb42, line 31: Is this alerting daily?

neb42 (commented):
Does smart alert give us alerts when the app goes down? As a side thought, how would azure know this? We don't have any sort of health check endpoint.

I don't see any alerts for cpu usage. And for memory it's just for leaks. What about sustained high cpu/memory usage indicating that we need more resources?

Does any of the alerts cover the database?

I'd be happy to fix my comment about the terraform variable, merge this, then add any other alerts in subsequent prs.

deploy/terraform/azurerm_monitor_smart_detector.tf
internetti: https://learn.microsoft.com/en-us/azure/azure-monitor/best-practices-alerts#alert-severity

deploy/terraform/azurerm_monitor_smart_detector.tf
internetti: this is once a day, yes. these values come from the ISO 8601 format https://tc39.es/proposal-temporal/docs/duration.html#:~:text=Briefly%2C%20the%20ISO%208601%20notation,suffix%20that%20indicates%20the%20unit.

internetti:
this article is what I used for the general structure of how to deploy the logic app:
https://azapril.dev/2021/04/12/deploying-a-logicapp-with-terraform/
creating an api_connection from azure to teams is apparently something between difficult and impossible, so it's easier to deploy the app with an arm template. the api connection "teams" has been manually created.

**current state of things:**

I added parameters for the threads to the app, but I still had to give proper values in the default values to be able to test the app via "run trigger"

I tried also having parameters for the team id and the channel id, but in the test run, I got an error about an invalid GUID, so I rolled that back.

You'll find the message, team and channel IDs in the link you get, when you do "copy link" on a teams message
this is an example: 
´´´
https://teams.microsoft.com/l/message/19:b89d9001773e4e9486792bca76534be0@thread.skype/1680101967978?tenantId=3f86d1b4-2c6b-4b1a-b2ae-6f5e950bca1e&groupId=d9e2211d-6b26-4b16-984a-36ae056c40ae&parentMessageId=1680101967978&teamName=HO%20Global%20ICT&channelName=ICT%20Service%20Alerts%20(WIP)&createdTime=1680101967978&allowXTenantAccess=false
´´´
19:b89d9001773e4e9486792bca76534be0@thread.skype is the channel id, the groupId d9e2211d-6b26-4b16-984a-36ae056c40ae is the team id and the parentMessageId 1680101967978 is the message id

when running the app with "run trigger" and the sample payload from the attached file, it sent the alerts to the prod messages, despite the logic app living in dev.
there is a "prod" in the URL of the test run, might be related 
(https://prod-183.westeurope.logic.azure.com:443/workflows/25f7ea987f4749438414b6bc9c5bbabd/triggers/manual/paths/invoke?api-version=2016-10-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig=EfpXJYxHh6W4PR8x_aC12ts2Lch4W8UaW5LjQEotkDc)

the azurerm_management_lock resource is not being automatically redeployed after removed in the UI, I had to put that back up in the UI. I suspect the property count to be the problem, cause it's not in the docs, but removing it caused an error.

the api connection to teams might also have to be parametrized since it's bound to a resource group

[logicapp-samplepayload.txt](https://github.com/NorwegianRefugeeCouncil/core/files/11122892/logicapp-samplepayload.txt)

deploy/terraform/azurerm_monitor_metric_alert.tf
```diff
@@ -0,0 +1,198 @@
...
+  description         = "Action will be triggered when the CPU percentage average is greater than 80."
+  frequency           = "PT1M"
+  severity            = 3
+
+  criteria {
+    metric_namespace = "Microsoft.DBforPostgreSQL/flexibleServers"
+    metric_name      = "cpu_percent"
+    aggregation      = "Average"
+    operator         = "GreaterThan"
+    threshold        = 80
```
internetti, line 23: 80 seems to be not enough, should be updated to 90

deploy/terraform/azurerm_monitor_action_group.tf
```diff
@@ -0,0 +1,12 @@
+resource "azurerm_monitor_action_group" "ag_teams" {
+  name                = "send-notification-to-teams-${var.environment}"
+  provider            = azurerm.runtime
+  resource_group_name = azurerm_resource_group.rg.name
+  short_name          = "notify-teams"
+
+  webhook_receiver {
+    name                    = "send_alert_to_okta_workflows"
+    service_uri             = "https://nrc.workflows.oktapreview.com/api/flo/bb459c9dcfe01856889a5882639cdea2/invoke?clientToken=fdbde6450a07740538547a2749139ea1703bd5f84d44bcee5b9ca33885f9f530"
```
neb42, line 9: This should be a variable

deploy/terraform/variables.tf
```diff
@@ -323,4 +323,9 @@ variable "frontdoor_sku_name" {
   type        = string
   description = "The sku name of the frontdoor."
   default     = "Standard_AzureFrontDoor"
+}
+
+variable "action_group_webhook_url" {
```
neb42, line 328: I don't see this being used. Is this meant to be be for the okta url?

internal/server/routes.go
```diff
@@ -41,8 +41,9 @@ func buildRouter(
 	staticRouter := r.PathPrefix("/static").Subrouter()
 	staticRouter.HandleFunc("/{file:.*}", web.ServeStatic)
 
-	webRouter := r.PathPrefix("").Subrouter()
+	r.Path("/health").Handler(handlers.HandleHealth())
```
neb42, line 44: Super minor but I think convention is `healthz` as it avoids conflict with any potential real name resources

neb42 (commented):
The health endpoint still needs to be configured in the web app.

I think in this file: `deploy/terraform/azurerm_linux_web_app.app.tf`
You need to add these fields:
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#health_check_path
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#health_check_eviction_time_in_min

This may be something to add too? https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#x_fd_health_probe

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app

deploy/terraform/azurerm_monitor_action_group.tf
internetti: oh yeah, I had even already put that into devops azure. forgot to use it ^^'

deploy/terraform/variables.tf
internetti: exactly ^^

neb42:
* Change cpu metric to cpu percentage on app service plan
* Change memory alert to higher value, try as percentage if possible
* Increase postgres resources

deploy/terraform/azurerm_monitor_metric_alert.tf
```diff
@@ -0,0 +1,205 @@
...
+  severity            = 2
+  frequency           = "PT1H"
+  window_size         = "PT1H"
+
+  criteria {
+    metric_namespace = "Microsoft.Web/serverfarms"
+    metric_name      = "CpuPercentage"
+    aggregation      = "Average"
+    operator         = "GreaterThan"
+    threshold        = 15
```
neb42, line 75: I think 15 is way too low. Maybe 70%?

deploy/terraform/azurerm_monitor_metric_alert.tf
```diff
@@ -0,0 +1,205 @@
...
+  severity            = 2
+  frequency           = "PT1H"
+  window_size         = "PT1H"
+
+  criteria {
+    metric_namespace = "Microsoft.Web/serverfarms"
+    metric_name      = "MemoryPercentage"
+    aggregation      = "Average"
+    operator         = "GreaterThan"
+    threshold        = 40
```
neb42, line 99: Again probably way too low

deploy/terraform/azurerm_monitor_metric_alert.tf
```diff
@@ -0,0 +1,205 @@
...
+# monitor web app
+# metric_names https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported#microsoftwebsites
+#################
+resource "azurerm_monitor_metric_alert" "app_health_check" {
+  provider            = azurerm.runtime
+  name                = "health-check-${var.environment}"
+  resource_group_name = azurerm_resource_group.rg.name
+  scopes              = [azurerm_linux_web_app.app.id]
+  description         = "${var.environment} - Web App: HealthCheckStatus is less than 100% okay."
+  severity            = 1
```
neb42, line 118: What do the different severity levels mean?

azure-pipelines.yml
neb42: Reminder to update this

neb42 (commented):
I think this is looking good now. Probably just worth updating the values to something more reasonable.

It'd be good to merge this soon so I can deploy it with the pen test deployment.

deploy/terraform/azurerm_monitor_metric_alert.tf
internetti: https://learn.microsoft.com/en-us/azure/azure-monitor/best-practices-alerts#alert-severity

PR #219, opened 2023-05-09
Static text localisation

internetti:


neb42:
Change the debug `es.toml` to be japanese or something

neb42 (approved):

PR #221, opened 2023-05-16
Switch language

internetti:
Added a server flag to enable the feature on localhost


internal/locales/jp.toml
```diff
@@ -1,3 +1,5 @@
+language = "es"
```
neb42, line 1: jp

internal/locales/jp.toml
```diff
@@ -137,11 +139,10 @@ warning_cannot_be_undone = "XXXX"
...
 download_template = "XXXX"
 edit_countries = "XXXX"
-files = "XXXX"
-logout = "XXXX"
+files = "Archivos"
 navigating_away = "XXXX"
-participants = "XXXX"
-search = "XXXX"
+participants = "Participantes"
+search = "Buscar"
```
neb42, line 145: XXX

neb42, line 145: Why does this file have changes but not the en file?

internal/server/middleware/localize.go
```diff
@@ -1,26 +1,46 @@
...
 import (
+	"github.com/nrc-no/notcore/internal/containers"
 	"net/http"
+	"strings"
 
 	"github.com/nicksnyder/go-i18n/v2/i18n"
 	"github.com/nrc-no/notcore/internal/locales"
 )
 
+var cookieName = "nrc-core-language"
```
neb42, line 12: const

internal/server/middleware/localize.go
```diff
@@ -1,26 +1,46 @@
...
+			language = l
+			break
+		}
+	}
+
+	if cookie != nil && availableLangs.Contains(cookie.Value) {
+		language = cookie.Value
+	}
+	return language
+}
```
neb42, line 46:
```
func getAppropriateLanguage(languageHeader string, cookie *http.Cookie, availableLangs containers.StringSet) string {
	if cookie != nil && availableLangs.Contains(cookie.Value) {
		return cookie.Value
	}

	headerLanguages := []string{}
	for _, lang := range strings.Split(languageHeader, ",") {
		headerLanguages = append(headerLanguages, strings.Split(lang, ";")[0])
	}
	for _, l := range headerLanguages {
		if availableLangs.Contains(l) {
			return l
		}
	}

	return locales.DefaultLang.String()
}
```

web/templates/nav.gohtml
```diff
@@ -146,6 +154,22 @@
...
+                                    aria-expanded="false">
+                                <i class="bi bi-globe"></i>
+                                {{.RequestContext.Locales.Translate "language"}}
+                            </button>
+                            <div class="dropdown-menu dropdown-menu-end">
+                                {{range $li, $l := .RequestContext.Locales.GetAvailableLangs}}
+                                    <button class="dropdown-item" onclick="setLanguageCookie({{$l}})">{{$l}}</button>
+                                {{end}}
+                            </div>
+                        </div>
```
neb42, line 172: It would be better if this was "English, Español, Français" rather than "en, es, fr"

internal/handlers/countries.go
```diff
@@ -49,6 +49,6 @@ func HandleCountries(templates map[string]*template.Template) http.Handler {
 
 		renderView(templates, templateName, w, r, viewParams{
 			viewParamsAllowedCountries: allowedCountries,
-		})
+		}, "")
```
neb42, line 52: I think we should actually pass the flag in here so that if someone adds a beta feature to this view it already works

internal/handlers/country.go
```diff
@@ -39,7 +39,7 @@ func HandleCountry(templates map[string]*template.Template, repo db.CountryRepo)
 		render := func() {
 			renderView(templates, templateName, w, r, viewParams{
 				viewParamCountry: country,
-			})
+			}, "")
```
neb42, line 42: same here

internal/handlers/individual.go
```diff
@@ -49,7 +49,7 @@ func HandleIndividual(templates map[string]*template.Template, repo db.Individua
 				"form":              individualForm,
 				"Individual":        individual,
 				templateParamAlerts: alerts,
-			})
+			}, "")
```
neb42, line 52: same here

neb42 (commented):
I think the flag should be a boolean like `--enable-beta-features` rather than an env. As we'd like to be able to test it locally with either settings.

The flag also needs to be set in the make file command.

internal/handlers/countries.go
internetti: I ended up doing that in the next PR ^^ I'll move it here

Makefile
```diff
@@ -69,7 +69,8 @@ serve: prerequisites
 		--hash-key-1="583ee90ea5821cb8c49ac0a8feeb7d090d597b1502fe653d68918122247d9675" \
 		--block-key-1="1e4254ae1f74406b853193466816ddffd592f8ef3bf9c2299735aec453a34239" \
 		--hash-key-2="28382c602c786507d5a286ff5890f1db12dc638355f50ab7cf91a946b07da670" \
-		--block-key-2="f735998d3da9d8a2b01cba8e9243605a7e02dd5daebefe62e23f2bc0b29787c2"
+		--block-key-2="f735998d3da9d8a2b01cba8e9243605a7e02dd5daebefe62e23f2bc0b29787c2" \
+		--enable-beta-features="true"
```
neb42, line 73: Normally with boolean flags just `--enable-beta-features` should evaluate to true and not passing the flag is false

Makefile
```diff
@@ -69,7 +69,8 @@ serve: prerequisites
 		--hash-key-1="583ee90ea5821cb8c49ac0a8feeb7d090d597b1502fe653d68918122247d9675" \
 		--block-key-1="1e4254ae1f74406b853193466816ddffd592f8ef3bf9c2299735aec453a34239" \
 		--hash-key-2="28382c602c786507d5a286ff5890f1db12dc638355f50ab7cf91a946b07da670" \
-		--block-key-2="f735998d3da9d8a2b01cba8e9243605a7e02dd5daebefe62e23f2bc0b29787c2"
+		--block-key-2="f735998d3da9d8a2b01cba8e9243605a7e02dd5daebefe62e23f2bc0b29787c2" \
+		--enable-beta-features
```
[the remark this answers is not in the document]
neb42: I think `len(flagValue)` would equal 0
```
func getFlagOrEnv(cmd *cobra.Command, flagName string, envName string) string {
	flagValue := cmd.Flag(flagName).Value.String()
	if len(flagValue) > 0 {
		return flagValue
	}
	return os.Getenv(envName)
}
```

neb42 (approved):

PR #222, opened 2023-05-17
translate form

internetti:
translating the individual form.

exporting the translator function so it can be used anywhere in the code.
added it to the template functions and removed it from the request context.

- relevant files to look at:
  - for general setup:
    - internal/locales/i18n.go
    - internal/server/middleware/localize.go
    - internal/server/routes.go
    - internal/server/templates.go
    - pkg/views/forms/render.go
  - for usage:
    - internal/views/individual_form.go
    - web/templates/countries.gohtml 
    - internal/api/enumTypes/disability_level.go


internal/locales/i18n.go
```diff
@@ -23,13 +22,22 @@ var localeFiles = map[string]string{
...
 var Translations *i18n.Bundle
 var DefaultLocalizer *i18n.Localizer
+var l *locales
 
-var AvailableLangs = containers.NewStringSet()
+type locales struct {
+	localizer *i18n.Localizer
+}
+
+func New() {
```
neb42, line 37: I would rename this to `Init` or `Initialise`. `New` indicates that it returns an instance of the class.

internal/api/enumTypes/enumTypes_test.go
neb42, line 1: Is this file doing anything?

internal/views/individual_form.go
```diff
@@ -176,9 +177,9 @@ func (f *IndividualForm) build() error {
 
 func (f *IndividualForm) buildTitle() error {
 	if f.isNew() {
-		f.Form.Title = "New Participant"
-	} else if f.individual.FullName == "" && f.individual.FirstName == "" && f.individual.MiddleName == "" && f.individual.LastName == "" && f.individual.PreferredName == "" && f.individual.NativeName == "" {
-		f.Form.Title = "Anonymous Participant"
+		f.Form.Title = f.Locales.Translate("new_participant")
+	} else if f.individual.FullName == "" {
```
neb42, line 181: Why have you removed several of the conditions?

internal/views/individual_form.go
```diff
@@ -904,10 +905,10 @@ func (f *IndividualForm) buildServiceCC(idx int) func() error {
 		}
 
 		options := getServiceCCOptions()
-		options = append([]forms.SelectInputFieldOption{{Value: "", Label: "Select a value"}}, options...)
+		options = append([]forms.SelectInputFieldOption{{Value: "", Label: f.Locales.Translate("select_a_value")}}, options...)
 		return buildField(&forms.SelectInputField{
 			Name:        fmt.Sprintf("service_cc_%d", idx),
-			DisplayName: fmt.Sprintf("Type of service #%d", idx),
+			DisplayName: f.Locales.Translate(fmt.Sprintf("service_cc_%d", idx)),
```
neb42, line 911: Do we want to translate the number here? Or just do:
```
fmt.Sprintf("%s #%d", f.Locales.Translate("service_cc"), idx)
```
Same for similar instances below.

web/templates/nav.gohtml
```diff
@@ -36,6 +36,10 @@
         }
 
         document.addEventListener("DOMContentLoaded", function () {
+            const uploadButton = document.getElementById(uploadIndividualFileButtonId)
+            if (uploadButton) {
+                uploadButton.addEventListener('click', openUploadDialog);
+            }
```
neb42, line 42: I'm a bit confused as to why this is here. Doesn't seem related to the rest of the pr and the upload buttons seems to work fine?

neb42:
Donwnloading and uploading the template file gives this error
image

neb42 (approved):

web/templates/nav.gohtml
internetti: no idea either, looks like a rebase issue. deleted.

internal/views/individual_form.go
internetti: I vaguely remember having a reason, but I cannot remember it

internal/api/enumTypes/enumTypes_test.go
internetti: yes, this initialises the locales for the enum tests, the naming of the functions follows conventions by which our testing framework knows what to do with it

internal/api/individual_tabular.go
```diff
@@ -906,6 +906,12 @@ func (i *Individual) marshalTabularData() ([]string, error) {
 			row[j] = string(v)
 		case enumTypes.DisplacementStatus:
 			row[j] = string(v)
+		case enumTypes.EngagementContext:
```
internetti, line 909: this is why the template was broken

internal/api/individual_tabular_test.go
```diff
@@ -52,6 +53,8 @@ var parameters = []struct {
 }
 
 func TestUnmarshalIndividualsTabularData(t *testing.T) {
+	locales.Init()
```
internetti, line 56: the tests don't succeed when the locales is not initialized

internal/server/middleware/localize.go
```diff
@@ -22,9 +21,7 @@ func Localize(enableBetaFeatures bool) func(handler http.Handler) http.Handler {
 				language = getAppropriateLanguage(languageHeader, languageCookie, locales.AvailableLangs)
 			}
 
-			localizer := i18n.NewLocalizer(locales.Translations, language)
-			ctx = locales.WithLocalizer(ctx, localizer)
-			r = r.WithContext(ctx)
+			locales.SetLocalizer(i18n.NewLocalizer(locales.Translations, language))
```
internetti, line 24: not attaching the localizer to the context anymore, since localisation isn't context dependent

internal/locales/en.toml
```diff
@@ -26,133 +26,28 @@ cancel = "Cancel"
 confirm_delete_individual = "Are you sure you want to delete this participant?"
 create_new_individual = "New participant"
 deactivate_individual = "Deactivate"
-deactivated_individual = "Deactivated Participant"
+deactivated_individual = "Deactivated participant"
```
dfigueroa87, line 29: Some questions/comments/ideas about the translation files in general:

- How's the translation process going to be? Will we use any tool to ease the translator's work or just provide them with this file? Also, how will we be adding new keys and getting them translated afterwards?
We don't really need to make a decision right now, but I think those are good things to start thinking about, and will affect how we structure these files and what format we use. 
I think for starters it could make sense to group keys per screen/form where they are present. E.g. `individual.deactivated`.

- Will we need to support pluralization? If so, how would we format it in this file?

- Are `individual` and `participant` meaning the same? If so, can we use just one term for them? I think could be useful to do a little Domain-Driven Design exercise to define the system and build a glossary so everyone in the team can refer to things in a single way and avoid confusions later on.

internal/locales/ja.toml
```diff
@@ -1,6 +1,6 @@
-language = "日本"
-en = "English"
-ja = "日本"
+language = "XXXX"
+en = "XXXX"
```
dfigueroa87, line 2: Is this meant to be used to select the language? Not a big deal, but I think it makes sense to just have the language name in the corresponding language instead of translating it to the selected language. That helps testing so once you for example switch to Japanese, you can find the way to change back to English without knowing Japanese.
So when you go to the language selection, you'd see a list like these:
- English
- 日本語
- Español
- etc.

internal/locales/en.toml
internetti: * do you know a good tool? so far, it's just giving them this file. if the need is there, maybe we could add the editing of translations as a feature in the app in v2. 
about the domain driven approach, sounds good, would be better, I'm sure we'll have to fight about where to put some of them though ^^
* probably, we currently have the TranslateCount function to support that. we don't use it though ...
* yes, they mean the same, and they should be called the same. the term participant came later. I'mm a big friend of consistency, but I don't think it's worth changing this in V1.

dfigueroa87 (approved):
:raised_hands:

dfigueroa87 (approved):

internal/locales/en.toml
dfigueroa87: Sorry I just realised I didn't answer.
I used Crowdin in the past, but that's probably too much for our needs (and likely expensive too). I also used a terrible tool before that I won't even mention the name. But I think we can look for something that covers our needs when building this feature for v2.

PR #232, opened 2023-06-29
increase app instances & remove dns records

internetti:


deploy/terraform/azurerm_service_plan.plan.tf
```diff
@@ -5,4 +5,5 @@ resource "azurerm_service_plan" "plan" {
   resource_group_name = azurerm_resource_group.rg.name
   os_type             = "Linux"
   sku_name            = var.service_plan_sku_name
+  worker_count = 2
```
neb42, line 8: Can we do this in a separate pr? The dns record change is a little experimental so I think it's best to decouple them

neb42 (commented):
There's a variable that needs to be removed too

neb42 (approved):

PR #244, opened 2023-08-17
COR-775: add collection agent to search filters

internetti:


neb42:
This search field isn't persisted in the query string

neb42 (approved):

PR #246, opened 2023-08-28
Envoy docker

neb42:
Something I did on ICLA Duhok that I thought is worth bring over to CORE. The drive behind it was that envoy doesn't work on windows outside of a docker container, but it also simplifies development.

This moves the envoy stuff into the docker compose. When developing we now only need to run `make up` and `make serve`, `make proxy` is no longer needed.

I've also moved the certs directory so after merging this will require `make certs` to be ran. Also there's a new make command to generate the certs.


deploy/envoy.yaml
```diff
@@ -222,11 +220,11 @@ static_resources:
               - endpoint:
                   address:
                     socket_address:
-                      address: 127.0.0.1
+                      address: host.docker.internal
```
dfigueroa87, line 223: `host.docker.internal` is only available for Docker in Mac and Windows. So my Linux workaround was to add it as `extra_hosts` in the docker-compose file (lines 75 and 76).
@neb42 can you test if it still works for you in Mac when you're back? :pray:

dfigueroa87:
@neb42 @internetti anything against getting this one merged?

PR #253, opened 2023-09-19
CPT-204: deduplicate on birthdate

internetti:
should you find some issues with the deduplication, please check if they're birthdate specific. 

I found some unrelated issues, will create tickets for them

### DIFF
--- a/internal/db/individual.go
+++ b/internal/db/individual.go
@@ -67,7 +67,11 @@ func (i individualRepo) findDuplicatesInternal(ctx context.Context, tx *sqlx.Tx,
 }
 
 type ColumnArgsGroups = map[string][]string
-type OrTypeArgsGroups = map[deduplication.DeduplicationTypeName]ColumnArgsGroups
+type TypedColumnArgsGroups = struct {
+	Args    ColumnArgsGroups
+	SQLType string
+}
+type OrTypeArgsGroups = map[deduplication.DeduplicationTypeName]TypedColumnArgsGroups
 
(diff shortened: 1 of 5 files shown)


internal/db/individual.go
```diff
@@ -81,10 +85,14 @@ type QueryArgs struct {
...
-		  "id_2": ["ID1", "ID2", "ID3", "ID4", "ID5","ID6"],
-		  "id_3": ["ID1", "ID2", "ID3", "ID4", "ID5","ID6"],
+		"IDs":{
+			Args: {
+					  "id_1": ["ID1", "ID2", "ID3", "ID4", "ID5","ID6"],
+					  "id_2": ["ID1", "ID2", "ID3", "ID4", "ID5","ID6"],
+					  "id_3": ["ID1", "ID2", "ID3", "ID4", "ID5","ID6"],
+					},
+			},
+			"SQLType: "text",
```
neb42, line 95: Not neccessarily suggesting this change, but a more correct data structure would be something like:
```
"IDs": {
  "text_field_1": {
    "type": "text",
    "values": [...],
  },
  "date_field_1": {
    "type": "time",
    "values": [...],
  },
}
```
as this would allow you to do comparisons against different data types

neb42 (approved):

internal/db/individual.go
internetti: I am not completely sure what you mean by that. Where exactly would this used to compare what with what?

PR #256, opened 2023-09-29
Cpt 2  Change authorization to use granular okta groups

neb42:
**DO NOT MERGE**
Requires Okta groups migration and a planned release

Before deploying:
* Ensure users have both the old okta groups and the new okta groups

After deploying:
* Go through each country and set their new okta groups in the configuration (users will not have access during this)

Changes:
* Remove nrc_organizations from countries, server options, and auth/permissions middleware
* Add read_group to countries
* Add write_group to countries
* Validate permissions from okta group and countries read/write group fields
* Add unittest make command
* Update local oidc users to provide clear set of test cases


internal/db/migrations/postgres/033_add_country_read_write_groups.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE countries
+    DROP COLUMN IF EXISTS nrc_organisations;
+
+ALTER TABLE countries
+    ADD COLUMN IF NOT EXISTS read_group VARCHAR(255) NOT NULL DEFAULT 'xxx';
```
internetti, line 5: is the xxx meant as a precaution? is there a scenario in which we'd get an empty string in the okta groups of a user?

neb42: Not why I did it but I guess to you help avoid that potentially happening. Either way it should be a very short lived state.

internetti (changes requested):
the country dropdown doesn't show countries with only read permissions.
you need to change line 120 in nav.gohtml to 
                                        {{if $auth.HasCountryPermissionRead $c.ID}}
or 
                                        {{if $auth.HasCountryPermissionRead $c.ID 2}}

looks like it was broken before


Also, what do you think about removing the Code field here?

internetti (approved):

PR #257, opened 2023-10-04
CPT-203: move vulnerabilities to own section & add comment field

internetti:


internal/db/sql_individual_list.go
```diff
@@ -730,6 +731,14 @@ func (g *getAllIndividualsSQLQuery) withPWDComments(pwdComments string) *getAllI
 	return g
 }
 
+func (g *getAllIndividualsSQLQuery) withVulnerabilityComments(vulnerabilityComments string) *getAllIndividualsSQLQuery {
+	if len(vulnerabilityComments) == 0 {
+		return g
+	}
+	g.writeString(" AND vulnerability_comments = ").writeArg(vulnerabilityComments)
```
neb42, line 738: Would `LIKE %$1%` make more sense here?

internetti: sure, why not. would also add it for the other comment fields then

neb42 (approved):

PR #258, opened 2023-10-17
use fixed terraform version

internetti:


Sripriyamenon88 (approved):

deploy/terraform/main.tf
```diff
@@ -15,6 +15,7 @@ terraform {
   }
   backend "azurerm" {
   }
+  required_version = "1.5.7"
```
Sripriyamenon88, line 18: we might want to confirm that this doesn't break any existing terraform resources due to deprecated functions/features

internetti: @Sripriyamenon88 this is the version that I found in the core-dev.tfstate file, so we've been using this version before, right?

PR #261, opened 2023-10-20
CPT-693: refactor deduplication

internetti:
Rewrite deduplication of database with query as suggested by Marta

Example:
when deduplicating on birth date, email and names, with All as the connecting condition

Step 1: 
Create a temporary table in which we store the individuals from the uploaded file. The table only contains the relevant columns (e.g. the deduplication columns plus id) to save memory. The table will be removed after the session. To prevent a mix up, the table name contains the request id.
`
CREATE TEMPORARY TABLE temp_individuals_<requestId>
(birth_date date,first_name varchar,middle_name varchar,last_name varchar,native_name varchar,email_1 varchar,email_2 varchar,email_3 varchar) 
ON COMMIT DROP;
`

Step 2:
Populate the temporary table with the individuals from the file.
`
INSERT INTO temp_individuals_<requestId> 
SELECT * FROM UNNEST($1::date[],$2::varchar[],$3::varchar[],$4::varchar[],$5::varchar[],$6::varchar[],$7::varchar[],$8::varchar[]);
`

Step 3:
Join the temporary table and the individuals table and query on the deduplication criteria.
`
SELECT DISTINCT 
ir.birth_date,ir.email_1,ir.email_2,ir.email_3,ir.first_name,ir.middle_name,ir.last_name,ir.native_name,ir.id 
FROM individual_registrations ir 
CROSS JOIN temp_individuals_<requestId> ti 
WHERE ir.country_id = $1 
AND ir.deleted_at IS NULL 
AND (ti.birth_date = ir.birth_date) 
AND (ti.email_1 = ir.email_1 OR ti.email_2 = ir.email_2 OR ti.email_3 = ir.email_3) 
AND (ti.first_name = ir.first_name AND ti.middle_name = ir.middle_name AND ti.last_name = ir.last_name AND ti.native_name = ir.native_name) 
AND ir.deleted_at IS NULL;
`

Other things I refactored
* combined deduplication types and operator to a config struct, cleaner function signatures, also, directly using the deduplication types, removing the need for lookups in the types list
* using df.FilterAggregation(dataframe.Or, filters...) for a clearer reading of how the filters are combined


internal/db/individual.go
```diff
@@ -27,241 +29,159 @@ type IndividualRepo interface {
...
 	if len(subQueries) > 0 {
-		subQuery := fmt.Sprintf("(%s)", strings.Join(subQueries, ") OR ("))
-		return subQuery, args
+		b.WriteString(fmt.Sprintf(" AND (%s)", strings.Join(subQueries, fmt.Sprintf(") %s (", config.Operator))))
+	}
+	if uploadDfHasIdColumn {
+		b.WriteString(" AND ti.id::uuid NOT IN (SELECT id FROM individual_registrations)")
 	}
-	return "", args
+	b.WriteString(" AND ir.deleted_at IS NULL;")
```
internetti, line 182: I don't fully understand why, but this condition needed to be doubled, ideas for fixes are welcome

internal/api/individual_deduplication.go
```diff
@@ -1,20 +1,21 @@
...
 	"github.com/nrc-no/notcore/pkg/api/deduplication"
+	"golang.org/x/exp/slices"
 	"strings"
 	"time"
 )
 
 func FindDuplicatesInUUIDColumn(df dataframe.DataFrame) []FileError {
 	filteredDf := df.Select([]string{indexColumnName, constants.FileColumnIndividualID, constants.FileColumnIndividualLastName})
 	fileErrors := []FileError{}
+	df.Col(indexColumnName).Records()
```
internetti, line 18: @internetti remove

internal/api/individual_deduplication.go
```diff
@@ -151,13 +143,19 @@ func getOrDuplicationScore(scoresByType []int, totalScores []int, df dataframe.D
...
-			result = df.Filter(filters...)
+		}
+		result := df.FilterAggregation(dataframe.Or, filters...)
 
-			for r := 0; r < result.Nrow(); r++ {
-				index, err := result.Select(indexColumnName).Elem(r, 0).Int()
-				if err == nil {
-					scoresByType[index]++
-				}
+		result = result.Filter(dataframe.F{
```
internetti, line 149: filter out the results that have the same id

internal/api/individual_deduplication.go
```diff
@@ -246,25 +267,35 @@ func FormatDbDeduplicationErrors(duplicates []*Individual, deduplicationTypes []
 					scores[rowNumber] = scores[rowNumber] + 1
 				}
 			}
+			if config.Operator == deduplication.LOGICAL_OPERATOR_OR {
+				if filteredDf.Nrow() > 0 {
+					break
```
internetti, line 272: if we are deduplicating with "Any" (=OR), then we can stop looking at the first matching condition

internal/api/individual_deduplication.go
```diff
@@ -326,3 +357,62 @@ func FormatFileDeduplicationErrors(duplicateMap []containers.Set[int], deduplica
...
+
+func GetDataframeFromRecords(records [][]string, deduplicationTypes []deduplication.DeduplicationType, mandatory []string) (dataframe.DataFrame, error) {
+	columnsOfInterest := []string{}
+	if len(deduplicationTypes) == 0 && len(mandatory) == 0 {
+		return dataframe.DataFrame{}, nil
+	}
+	for _, deduplicationType := range deduplicationTypes {
+		columnsOfInterest = append(columnsOfInterest, deduplicationType.Config.Columns...)
+	}
+	for _, mandatoryColumn := range mandatory {
```
internetti, line 369: mandatory like id to check if the duplicate is not actually a duplicate but needs to be updated, and last_name for the error messages

internal/db/individual.go
```diff
@@ -27,241 +29,172 @@ type IndividualRepo interface {
...
+						if err != nil {
+							values[i] = nil
+							continue
+						}
+						values[i] = &t
 					}
-					colType = "date"
+					args = append(args, pq.Array(values))
+				} else {
+					args = append(args, pq.Array(g.Records()))
```
internetti, line 157: for now we only deduplicate on dates or strings, might have to be asjusted later

dfigueroa87 (commented):
Looks good to me. Just added a minor comment.

In addition to that, I think there may be room for further refactoring in the handlers code (I particularly checked the `individuals_upload` handler). I think would be good to extract some logic into separate functions, and likely move business logic out of there. Maybe also change how error handling goes.
But probably not worth the effort at this stage.

dfigueroa87 (approved):

dfigueroa87:
Got into the branch to test, and I'm likely misusing the application, but I created a participant with just a full name and then tried to create a new one with the same full name selecting the deduplication on that field, which resulted in this error:
image

internetti:
> Got into the branch to test, and I'm likely misusing the application, but I created a participant with just a full name and then tried to create a new one with the same full name selecting the deduplication on that field, which resulted in this error: image

that does not sound like misusing the app, I'll check it out

dfigueroa87 (approved):

PR #274, opened 2023-12-07
CPT-401: file localization

internetti:
make template for all available languages

internally only use DBColumn... fields, FileColumn... fields only for translation purposes

translate columns back to internal fields as soon as possible, then keep working with the internal fields


internal/api/utils.go
```diff
@@ -107,7 +115,14 @@ func stringArrayToInterfaceArray(row []string) []interface{} {
 var indexColumnName = "index"
 
 func GetDataframeFromRecords(records [][]string) dataframe.DataFrame {
+	keys := locales.GetTranslationKeys(records[0])
+	dbCols := make([]string, len(keys))
+	for i, key := range keys {
+		dbCols[i] = constants.IndividualFileToDBMap[key]
+	}
+
 	return dataframe.LoadRecords(records,
```
internetti, line 124: this can be nicer, but I'll have to rework this anyway when rebasing the deduplication refactor/or the other way around

dfigueroa87 (approved):

PR #289, opened 2023-12-19
CPT-402: RTL

internetti:
* added debug file for rtl language
* add scipt for translating the css theme into rtl version, so we have two css files, one rtl and one ltr
* add isRTL variable in templates to set corresponding theme css and left/right values for sticky columns in the table


web/static/theme.min.rtl.css
```diff
@@ -0,0 +1,6 @@
+/*!
```
dfigueroa87, line 1: I don't think this file should be committed. The other `theme.min` files are not.

dfigueroa87 (commented):
Looks good, just added one comment

dfigueroa87 (approved):

PR #290, opened 2023-12-19
feat(downloads): Handle file downloads using Azure blob storage

dfigueroa87:
Improvement for downloading files that uploads them to Azure blob storage and then streams them from there through the same download API endpoint. 
The files were previously stored in the server instances, not allowing us to properly scale instances up.

The Azure storage setup uses private links that are not generally available in the public internet.


deploy/terraform/azurerm_storage_account.download_storage.tf
```diff
@@ -0,0 +1,59 @@
+resource "azurerm_storage_account" "download_storage" {
```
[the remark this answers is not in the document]
dfigueroa87: Fixed in https://github.com/NorwegianRefugeeCouncil/core/pull/290/commits/b2c8b7926626dcf82d59e8678dd2db56c1b0e52e

dfigueroa87:
This PR is introducing a new security hotspot since it doesn't enable logging for the Azure blob storage.
Unfortunately terraform doesn't allow to enable it programatically, so as far as I understand we would need to do it manually.
@neb42 / @internetti any thoughts or suggestions?

build/package/app.Dockerfile
```diff
@@ -10,13 +10,14 @@ RUN apt-get install -yq --no-install-recommends curl && \
     npm install -g yarn
 ARG uid=1000
 ARG gid=1000
-RUN groupadd -g ${GROUP_ID} core || echo "Group already exists"
-RUN useradd -l -m -d /home/app -u ${USER_ID} -g ${GROUP_ID} core || echo "User already exists"
+RUN groupadd -g ${gid} core || echo "Group already exists"
```
dfigueroa87, line 13: The changes in this file are not really related to the PR, but for some reason the CI pipelines started to fail and I made these changes to fix it.

deploy/docker-compose.yaml
```diff
@@ -73,4 +73,15 @@ services:
...
+    image: mcr.microsoft.com/azure-storage/azurite
+    container_name: individualdownloads
+    command: 'azurite -l /data --blobHost 0.0.0.0 --blobPort 11000 --queueHost 0.0.0.0 --queuePort 11001 --tableHost 0.0.0.0 --tablePort 11002'
+    ports:
+      - '11000:11000'
+      - '11001:11001'
+      - '11002:11002'
+    environment:
+      - ACCOUNT_NAME=devstoreaccount1
+      - ACCOUNT_KEY="Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
```
internetti, line 87: can we put those in the github variables/secrets?

internal/handlers/individuals_download.go
```diff
@@ -188,19 +208,36 @@ func HandleDownload(
...
 			selectedCountryID,
-			path.Base(downloadFile.Name()),
+			path.Base(fileName),
 		)
 		http.Redirect(w, r, redirectPath, http.StatusSeeOther)
-
 	})
 }
+
+func parseRangeHeader(rangeHeader string) (int64, int64, error) {
```
internetti, line 226: a test for this one would be nice

internal/server/server.go
```diff
@@ -166,3 +177,38 @@ func (s *Server) Start(ctx context.Context) error {
...
 	return nil
 }
+
+func getAzureBlobStorageClient(ctx context.Context, url string)  (*azblob.Client, error) {
+	isDev := strings.Contains(url, "localhost") || strings.Contains(url, "127.0.0.1")
+
+	if isDev {
+		credential, err := azblob.NewSharedKeyCredential(
+			"devstoreaccount1",
+			"Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==",
```
internetti, line 187: these should go into a variable

deploy/docker-compose.yaml
dfigueroa87: Actually this is not needed here, this is the `well-known storage account` that azurite already knows about and we only need these values to connect

internal/server/server.go
dfigueroa87: These are the [well-known storage account](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=docker-hub%2Cblob-storage#well-known-storage-account-and-key) details defined by Azurite, that we only use for emulating azure storage when running the app locally, so I don't think we need to have it in a more global config file.
But I'll move it to a local variable here, to make it more clear

internal/handlers/individuals_download.go
```diff
@@ -188,19 +208,36 @@ func HandleDownload(
...
-			if err := os.Remove(downloadFile.Name()); err != nil {
-				l.Error("failed to remove temp file", zap.Error(err))
-			}
-		}()
+		_, err = azureStorageClient.UploadFile(ctx, containerName, fileName, downloadFile, nil)
+		if err != nil {
+			l.Error("failed to upload file", zap.Error(err))
+			http.Error(w, "failed to upload file: "+err.Error(), http.StatusInternalServerError)
+			return
+		}
```
internetti, line 216: why are you uploading the file here?

internal/server/server.go
```diff
@@ -166,3 +177,38 @@ func (s *Server) Start(ctx context.Context) error {
...
+		if err != nil {
+			return nil, err
+		}
+
+		client, err := azblob.NewClientWithSharedKeyCredential(url, credential, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		client.CreateContainer(ctx, "individualdownloads", nil)
```
internetti, line 198: can we get this name from the server options?

internal/handlers/individuals_download.go
dfigueroa87: The file generated to be downloaded is uploaded to Azure blob storage, and then a redirect is returned to the request, so the file can be downloaded from there through the same endpoint (that redirect will get into the `if` branch in line 95 since at that point the file in the query param is defined)

internal/server/server.go
```diff
@@ -166,3 +177,38 @@ func (s *Server) Start(ctx context.Context) error {
...
 	return nil
 }
+
+func getAzureBlobStorageClient(ctx context.Context, url string) (*azblob.Client, error) {
+	isLocalEnvironment := strings.Contains(url, "localhost") || strings.Contains(url, "127.0.0.1")
+
+	if isLocalEnvironment {
+		// we use Azurite to emulate Azure Blob Storage locally
+		wellKnownAzuriteAccountName := "devstoreaccount1"
+		wellKnownAzuriteAccountKey := "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
```
internetti, line 187: I personally would still prefer this to be in the makefile

dfigueroa87: There's two parts I don't like about it:
1. defining these values globally when they only need to be used for a particular case
2. breaking the pattern for generating the server options from flags and env vars, since in this case we'd only use flags as it's only relevant for the local environment emulating the blob storage

But it's not really critical, I'll make the change.\

One other though, related to the 2nd point. Is there a reason for using both flags and env variables? I'd argue that just using env variables would simplify the whole setup quite a bit, and the end result would be the same.

internal/server/server.go
```diff
@@ -166,3 +177,38 @@ func (s *Server) Start(ctx context.Context) error {
 	}
 	return nil
 }
+
+func getAzureBlobStorageClient(ctx context.Context, url string) (*azblob.Client, error) {
+	isLocalEnvironment := strings.Contains(url, "localhost") || strings.Contains(url, "127.0.0.1")
```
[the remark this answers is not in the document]
dfigueroa87: I generally prefer to move things to `utils` when they start being re-used, just to avoid duplication. 
But not big harm doing it here :)

internal/server/server.go
internetti: valid points
about the second one, I guess not? we could probably work only with env variables.

deploy/terraform/azurerm_subnet.runtime_subnet.tf
```diff
@@ -13,4 +13,34 @@ resource "azurerm_subnet" "runtime_subnet" {
...
+  name                = "runtime-nsg"
+  location            = azurerm_resource_group.rg.location
+  resource_group_name = azurerm_resource_group.rg.name
+}
+
+# Outbound traffic to the Azure Instance Metadata Service (IMDS) using its well-known IP address
+# This is required to use Managed Identities
+resource "azurerm_network_security_rule" "outbound_rule_imds" {
+  provider                    = azurerm.runtime
+  name                        = "allow-outbound"
```
neb42, line 29: "allow-outbound-azure-imds"

neb42 (commented):
On the deployed instance, downloading a large file (~10k rows) downloads a corrupted file. I suspect that small files include the entire file in the first chunk of the range header and when it tries to download in multiple chunks there is some sort of issue.

neb42:
> This PR is introducing a new security hotspot since it doesn't enable logging for the Azure blob storage. Unfortunately terraform doesn't allow to enable it programatically, so as far as I understand we would need to do it manually. @neb42 / @internetti any thoughts or suggestions?

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account

There's a logging block in the docs here.

dfigueroa87:
> > This PR is introducing a new security hotspot since it doesn't enable logging for the Azure blob storage. Unfortunately terraform doesn't allow to enable it programatically, so as far as I understand we would need to do it manually. @neb42 / @internetti any thoughts or suggestions?
> 
> https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account
> 
> There's a logging block in the docs here.

That logging block refers to the `queue_properties` block that allows that property within it. But `logging` is not an allowed property in the storage account itself.

neb42:
I downloaded the file from the azure ui and it was fine, so definitely the download logic.

neb42:
Looking at my downloaded files, they all seem to be 8.4mb when this error happens. I suspect this is the max chunk size from frontdoor and it isn't downloading the next chunk. I vaguely remember this happening before but I can't recall the details.

internal/handlers/individuals_download.go
```diff
@@ -87,36 +93,46 @@ func HandleDownload(
...
-				http.Error(w, "failed to open file: "+err.Error(), http.StatusInternalServerError)
+				l.Error("failed to read file", zap.Error(err))
+				http.Error(w, "failed to read file: "+err.Error(), http.StatusInternalServerError)
 				return
 			}
 
 			setContentTypeForExtension(w, resultFileExtension)
-			// serve the file
-			http.ServeContent(w, r, resultFileName, time.Time{}, file)
+			http.ServeContent(w, r, resultFileName, time.Time{}, bytes.NewReader(b))
```
neb42, line 135: I think `ServeContent` respects the range header. Seen as we handle the range header in the blob storage download then we can ignore it here. I think `ServeFile` will give us the behaviour we want.

Or we download the entire file on each chunk. Not ideal but maybe easier.

deploy/terraform/azurerm_private_dns_zone.storage_dns.tf
```diff
@@ -0,0 +1,5 @@
+resource "azurerm_private_dns_zone" "storage_dns" {
+  provider            = azurerm.runtime
+  name                = "privatelink.blob.core.windows.net"
```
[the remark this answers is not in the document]
dfigueroa87: This actually needs to be named `privatelink.blob.core.windows.net` to work properly with the configured private endpoint.

```
when using Azure Blob Storage with a private endpoint, it's a requirement to use a Private DNS Zone named privatelink.blob.core.windows.net.

When you create a private endpoint for your storage account, Azure creates a DNS record in a Private DNS Zone that maps the storage account's blob service endpoint (e.g., mystorageaccount.blob.core.windows.net) to the private IP address of the private endpoint. This DNS record allows clients in your virtual network to access the storage account using the private endpoint.
```

deploy/docker-compose.yaml
```diff
@@ -73,4 +73,12 @@ services:
       - ./certs/tls.key:/etc/envoy/certs/tls.key
       - ./certs/tls.crt:/etc/envoy/certs/tls.crt
     extra_hosts:
-      - "host.docker.internal:host-gateway"
\ No newline at end of file
+      - "host.docker.internal:host-gateway"
+  azure_blob_storage:
+    image: mcr.microsoft.com/azure-storage/azurite
+    container_name: individualdownloads
```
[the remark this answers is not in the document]
dfigueroa87: Good catch, that's just defining the docker container name, which we don't really need. I'll remove it.
The Azurite container is created when the server starts in a local environment.

internal/handlers/individuals_download.go
dfigueroa87: I opted for removing all the custom range handling logic, and letting `ServeContent` deal with it instead. This results in the whole file being retrieved from Azure Storage to the App Server for each chunk, which I agree is not ideal but couldn't make it work otherwise.
The latest version is deployed to DEV now, and I tested with big files which are downloaded correctly. Pausing and resuming the download from the browser is working fine too. Was there any other use case for having the custom range logic here?

azure-pipelines.yml
```diff
@@ -77,7 +77,7 @@ stages:
 
     - task: Docker@2
       displayName: Build final image
-      condition: and(succeeded(), eq(variables.IS_MAIN, 'true'))
+      condition: succeeded()
```
dfigueroa87, line 80: Changes to this file are just for allowing to make releases from the dev branch. To be undone before merging.

internal/handlers/individuals_download.go
```diff
@@ -87,36 +90,32 @@ func HandleDownload(
...
-				http.Error(w, "failed to get file info: "+err.Error(), http.StatusInternalServerError)
+				l.Error("failed to download file", zap.Error(err))
+				http.Error(w, "failed to download file: "+err.Error(), http.StatusInternalServerError)
 				return
 			}
+			defer downloadStream.Body.Close()
 
-			// open the file
-			file, err := os.Open(filePath)
+			data, err := io.ReadAll(downloadStream.Body)
```
dfigueroa87, line 109: This is downloading the whole stream/file to memory, which could be a problem for huge files / at scale. An alternative could be to put it to disk instead, which would add extra latency. 
@neb42 how big are we expecting these files to be? Should we go with this approach for now and evaluate it later if we run into issues?

neb42: I think we estimate the upper end to be ~200k rows. `go run . mock-data --count=200000` will generate a file to test that.

I'd say the extra latency is fine for this as I'd expect the network to be the bottle neck in most cases for large files.

dfigueroa87: A generated 200k rows xlsx file was ~220 MB. So I changed the implementation to persist a temporary file to disk instead of serving it from memory. The change is already deployed to DEV.
@neb42 please give it another look whenever you have some time.

internal/handlers/individuals_download.go
```diff
@@ -188,19 +222,35 @@ func HandleDownload(
...
-			path.Base(downloadFile.Name()),
+			path.Base(fileName),
 		)
 		http.Redirect(w, r, redirectPath, http.StatusSeeOther)
-
 	})
 }
+
+func SaveStreamToTempFile(stream io.ReadCloser) (*os.File, error) {
+	tmpfile, err := os.CreateTemp("/tmp", "*")
```
neb42, line 241: What is the `*` doing here? I'd expect this to be a random string

internal/handlers/individuals_download.go
```diff
@@ -188,19 +222,35 @@ func HandleDownload(
...
 			selectedCountryID,
-			path.Base(downloadFile.Name()),
+			path.Base(fileName),
 		)
 		http.Redirect(w, r, redirectPath, http.StatusSeeOther)
-
 	})
 }
+
+func SaveStreamToTempFile(stream io.ReadCloser) (*os.File, error) {
```
neb42, line 240: `saveStreamToTempFile` as this function should be private

internal/handlers/individuals_download.go
dfigueroa87: The second parameter is a prefix for the file name, the asterisk is replaced by a random string. I guess an empty string would work as well, I'll take a look

PR #295, opened 2024-01-23
fix(upload): CPT-709 use db column ids mapping for identifying columns in files

dfigueroa87:
The issue was that we were comparing the localised header of the columns in the file to the name of the columns in the database, which in this case were `ID` against `id`, resulting in that column not being considered in the deduplication process.

Additionally found another similar issue, that was causing the `FindDuplicatesInUUIDColumn` function to never be executed, since it was trying to find a file column header named `file_id`. So I fixed it using the same approach above.

### DIFF
--- a/internal/handlers/individuals_upload.go
+++ b/internal/handlers/individuals_upload.go
@@ -14,7 +14,6 @@ import (
 	"github.com/nrc-no/notcore/internal/utils"
 	"github.com/nrc-no/notcore/pkg/api/deduplication"
 	"go.uber.org/zap"
-	"golang.org/x/exp/slices"
 )
 
 var UPLOAD_LIMIT = 10000
@@ -91,7 +90,8 @@ func HandleUpload(renderer Renderer, individualRepo db.IndividualRepo) http.Hand
 		deduplicationConfig, err := deduplication.GetDeduplicationConfig(deduplicationTypes, deduplicationLogicOperator)
 
 		mandatoryColumns := []string{constants.DBColumnIndividualLastName}
-		if _, idColumnExists := colMapping[constants.DBColumnIndividualID]; idColumnExists {
+		var idColumnExistsInFile bool
+		if _, idColumnExistsInFile = colMapping[constants.DBColumnIndividualID]; idColumnExistsInFile {
 			mandatoryColumns = append(mandatoryColumns, constants.DBColumnIndividualID)
 		}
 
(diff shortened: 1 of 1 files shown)


neb42 (changes requested):
This works when just updating rows. But if I upload a file with 1 row for an update and 1 row for a create, it will throw the same false positive duplicate.

dfigueroa87:
> This works when just updating rows. But if I upload a file with 1 row for an update and 1 row for a create, it will throw the same false positive duplicate.

@neb42 I tested updating and adding in the same file and that doesn't seem to be a problem. 
What I found to be a problem though is the strategy for comparing emails, and potentially any other field that allows empty strings. This is because the cross join will consider blank emails as the same email. 
Is there a reason for using empty strings instead of nulls? I think that's that would be the best solution for this problem. The alternative is to change all deduplication queries to consider empty strings, which would make all queries more complicated (less readable / maintainable).

neb42:
> > This works when just updating rows. But if I upload a file with 1 row for an update and 1 row for a create, it will throw the same false positive duplicate.
> 
> @neb42 I tested updating and adding in the same file and that doesn't seem to be a problem. What I found to be a problem though is the strategy for comparing emails, and potentially any other field that allows empty strings. This is because the cross join will consider blank emails as the same email. Is there a reason for using empty strings instead of nulls in the database? I think that's that would be the best solution for this problem. The alternative is to change all deduplication queries to consider empty strings, which would make all queries more complicated (less readable / maintainable).
> 
> Another limitation I found in the deduplication strategy for emails is that at the moments it's just comparing with the same email field but not the others, i.e. `email_1` only against `email_1`, `email_2` against `email_2` and so on. So if adding a new participant with a second email that an existing participant has as first email, the deduplication check will allow it. Is that behaviour expected?

I think both changes make sense. But maybe check with Fix/Masa first.

dfigueroa87:
@neb42 I created https://norwegianrefugeecouncil.atlassian.net/browse/CPT-717 and https://norwegianrefugeecouncil.atlassian.net/browse/CPT-718 for those 2 issues, to keep changes smaller and more specific. So this PR just fixes the issue specified in https://norwegianrefugeecouncil.atlassian.net/browse/CPT-709

dfigueroa87:
@neb42 is it fine if I merge this one? I'm taking care of the other issues in the different tickets/PRs.

neb42 (approved):

PR #298, opened 2024-01-31
fix(deduplication): CPT-717 Conditionally ignore empty texts in deduplication fields

dfigueroa87:
https://norwegianrefugeecouncil.atlassian.net/browse/CPT-717

Changed the deduplication query builder, so empty strings are not considered.
Particularly, for OR subqueries we ignore them, and for AND subqueries we ignore them when all fields are empty (e.g. for anonymous participants)

I opted for updating the queries instead of defining columns as nullable, since the risk of messing other parts of the application is lower and the changes more contained. 
For v2 I'd rather suggest to work with null values and make a difference for what null vs empty means.

### DIFF
--- a/pkg/api/deduplication/deduplicationTypes.go
+++ b/pkg/api/deduplication/deduplicationTypes.go
@@ -27,9 +27,17 @@ const (
 	LOGICAL_OPERATOR_AND LogicOperator = "AND"
 )
 
+type DataType string
+
+const (
+	DataTypeString DataType = "string"
+	DataTypeDate   DataType = "date"
+)
+
 type DeduplicationTypeValue struct {
 	Columns   []string
 	Condition LogicOperator
+	Type      DataType // defined as a single value since all columns have the same type at the moment, change to array if needed
 }
 
 type DeduplicationType struct {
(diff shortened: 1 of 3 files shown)


pkg/api/deduplication/deduplicationTypes.go
```diff
@@ -51,6 +59,7 @@ var DeduplicationTypes = map[DeduplicationTypeName]DeduplicationType{
 		Config: DeduplicationTypeValue{
 			Columns:   []string{constants.DBColumnIndividualPhoneNumber1, constants.DBColumnIndividualPhoneNumber2, constants.DBColumnIndividualPhoneNumber3},
 			Condition: LOGICAL_OPERATOR_OR,
+			Type:      DataTypeString,
 		},
```
dfigueroa87, line 63: Now I'm actually thinking that this abstraction is not really needed. We could just define the SQL subquery here instead, and simplify the logic in the `buildDeduplicationQuery` function in `internal/db/individual.go` a lot.
I'll create another PR with that approach, so we can discuss it.

dfigueroa87:
Closing in favor of https://github.com/NorwegianRefugeeCouncil/core/pull/299

PR #299, opened 2024-01-31
fix(deduplication): CPT-717 Defined subqueries considering empty strings

dfigueroa87:
https://norwegianrefugeecouncil.atlassian.net/browse/CPT-717

Alternative solution to https://github.com/NorwegianRefugeeCouncil/core/pull/298 

Changed the deduplication query builder, so empty strings are not considered.
Particularly, for OR subqueries we ignore them, and for AND subqueries we ignore them when all fields are empty (e.g. for anonymous participants)

I opted for updating the queries instead of defining columns as nullable, since the risk of messing other parts of the application is lower and the changes more contained. 
For v2 I'd rather suggest to work with null values and make a difference for what null vs empty means.

This one defined the subqueries directly in the dedup configuration, simplifying the approach but increasing coupling.

Also extended it to consider edge cases.
- If ALL fields in the criteria are empty, two entries are not considered as the same.
- If some fields in the criteria are empty, those fields are ignored and the non-empty ones considered for deduplication.

### DIFF
--- a/internal/db/individual.go
+++ b/internal/db/individual.go
@@ -207,10 +207,10 @@ SELECT DISTINCT  ir.birth_date,ir.email_1,ir.email_2,ir.email_3,ir.first_name,ir
 	CROSS JOIN temp_individuals_<requestId> ti
 	WHERE ir.country_id = $1
 		AND ir.deleted_at IS NULL
-		AND (ti.birth_date = ir.birth_date)
-		AND ((ti.email_1 != '' AND ti.email_1 = ir.email_1) OR (ti.email_2 != '' AND ti.email_2 = ir.email_2) OR (ti.email_3 != '' AND ti.email_3 = ir.email_3))
-		AND (ti.first_name = ir.first_name AND ti.middle_name = ir.middle_name AND ti.last_name = ir.last_name
-			AND ti.native_name = ir.native_name AND (ti.first_name != '' OR ti.middle_name != '' OR ti.last_name != '' OR ti.native_name != ''));
+		AND ((ti.birth_date = ir.birth_date)
+		AND ((ti.email_1 != '' AND ti.email_1 = ir.email_1) OR (ti.email_2 != '' AND ti.email_2 = ir.email_2) OR (ti.email_3 != '' AND ti.email_3 = ir.email_3) OR (ti.email_1 = '' AND ti.email_2 = '' AND ti.email_3 =''))
+		AND (ti.first_name = ir.first_name AND ti.middle_name = ir.middle_name AND ti.last_name = ir.last_name AND ti.native_name = ir.native_name)
+		AND (ti.email_1 != '' OR ti.email_2 != '' OR ti.email_3 != '' OR ti.first_name != '' OR ti.middle_name != '' OR ti.last_name != '' OR ti.native_name != '')));
 */
 func buildDeduplicationQuery(tempTableName string, columnsOfInterest []string, config deduplication.DeduplicationConfig, uploadDfHasIdColumn bool, schema []DBColumn) string {
 	b := &strings.Builder{}
@@ -224,18 +224,20 @@ func buildDeduplicationQuery(tempTableName string, columnsOfInterest []string, c
 	b.WriteString(" WHERE ir.country_id = $1 AND ir.deleted_at IS NULL")
 
 	subQueries := []string{}
-	consideredColumnsAndOperator := []string{}
+	consideredStringColumnsAndOperator := []string{}
 	for _, dt := range config.Types {
 		if config.Operator == deduplication.LOGICAL_OPERATOR_AND {
 			subQueries = append(subQueries, dt.Config.QueryAnd)
-			consideredColumnsAndOperator = append(consideredColumnsAndOperator, dt.Config.Columns...)
+			if dt.Config.Type == deduplication.DataTypeString {
+				consideredStringColumnsAndOperator = append(consideredStringColumnsAndOperator, dt.Config.Columns...)
+			}
 		} else {
 			subQueries = append(subQueries, dt.Config.QueryOr)
 		}
 	}
-	if len(consideredColumnsAndOperator) > 0 {
+	if len(consideredStringColumnsAndOperator) > 0 {
 		emptyColumnChecks := []string{}
-		for _, column := range consideredColumnsAndOperator {
+		for _, column := range consideredStringColumnsAndOperator {
 			emptyColumnChecks = append(emptyColumnChecks, fmt.Sprintf("ti.%s != ''", column), fmt.Sprintf("ir.%s != ''", column))
 		}
 		notAllEmptyQuery := strings.Join(emptyColumnChecks, " OR ")
--- a/internal/db/individual.go
+++ b/internal/db/individual.go
@@ -224,23 +224,19 @@ func buildDeduplicationQuery(tempTableName string, columnsOfInterest []string, c
 	b.WriteString(" WHERE ir.country_id = $1 AND ir.deleted_at IS NULL")
 
 	subQueries := []string{}
-	consideredStringColumnsAndOperator := []string{}
+	notEmptyPartialChecks := []string{}
 	for _, dt := range config.Types {
 		if config.Operator == deduplication.LOGICAL_OPERATOR_AND {
 			subQueries = append(subQueries, dt.Config.QueryAnd)
-			if dt.Config.Type == deduplication.DataTypeString {
-				consideredStringColumnsAndOperator = append(consideredStringColumnsAndOperator, dt.Config.Columns...)
+			if dt.Config.QueryNotAllEmpty != "" {
+				notEmptyPartialChecks = append(notEmptyPartialChecks, dt.Config.QueryNotAllEmpty)
 			}
 		} else {
 			subQueries = append(subQueries, dt.Config.QueryOr)
 		}
 	}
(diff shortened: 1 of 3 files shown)


neb42:
I think I prefer this approach as it's easier to read, even if it does involve a bit more SQL code. But I'd like @internetti to have a look at this as she wrote it originally and it has some complexity.

pkg/api/deduplication/deduplicationTypes.go
```diff
@@ -119,7 +131,8 @@ var DeduplicationTypes = map[DeduplicationTypeName]DeduplicationType{
 		Config: DeduplicationTypeValue{
 			Columns:   []string{constants.DBColumnIndividualFullName},
 			Condition: LOGICAL_OPERATOR_OR,
-			Type:      DataTypeString,
+			Query: fmt.Sprintf("ti.%s = ir.%s",
```
internetti, line 134: why are we not excluding empty values here?

dfigueroa87: That's the next problem I think we need to solve related to deduplication. There's some criteria fields that (I think) are meant to be used in conjuction, and when that's the case it does matter a lot what's the logical operator at the higher level (whether the user selected Any or All, so we do AND or OR).
And that needs to be considered in the comparisons for the sub-queries. 
In short, for AND queries, a subquery needs to evaluate as `true` to be ignored, but for an OR query it needs to be `false`.

So my decision at this point was to consider the scenarios in which sub-queries would be used and take the more-likely ones to define the query. In this case, I think Full Name is more likely to be used in an `All`/`AND` query, therefore when the values are empty I preferred to return a `true` result for the sub-query.

dfigueroa87: I agree this needs to be discussed. But the requirements are complex and I think the only/best way is to do it incrementally. This PR is solving some existing issues, and leaving some other issues untouched. This query was already like this, now with the refactor the issue is more obvious, but the end result is exactly the same as it was (for this fullName field).
My intention is, after integrating these changes, do another round of testing and create tickets for any other issues found. Then we can re-iterate over the solution.

Sure, suppose we have 3 subqueries, P0, P1 and P2.
`P0 AND P1 AND P2` will be true only when the 3 are true, and false in any other case.
`P0 OR P1 OR P2` will be false only when the 3 are false, and true in any other case.
Now, when one of the values within a subquery is empty we want to ignore that piece of the query. Let's say P2 is related to this field, so when it's empty we want the query to just consider P0 and P1. To do this we want the queries to be:
`P0 AND P1 AND true`
`P0 OR P1 OR false`
So that way P2 doesn't have an impact in the main result, and only the non-empty fields do.

internetti: while I generally like this approach, I'm wondering if this complexity means that we shouldn't basically hardcode the queries

dfigueroa87: But what's the alternative? Building the queries dynamically makes it way more complex in my opinion, and more error prompt.

If we follow the approach in this PR, I think the next step would be to have two sub-queries per criteria, one for `AND` and another for `OR` queries. If we follow the other approach, it'll likely be having an `if` branching into two different strategies for building the subqueries in a generic way (one for `AND`, one for `OR`).

internetti: that sounds okay. my first idea was to make the Query property a function that takes the logic operator, which would have meant to have business logic in the config, didn't like that. your approach doesn't do that, so fine, let's go with this then

internetti (changes requested):
This doesn't seem to be working for the individual form in the app, this one's still flagging a duplicate for two empty full name fields

dfigueroa87:
> This doesn't seem to be working for the individual form in the app, this one's still flagging a duplicate for two empty full name fields

I'll take a look. What fields have you chosen for deduplication? If it's only full name, then the issue is likely what we discussed in the thread above.
But if you agree with this as the preferred approach (over https://github.com/NorwegianRefugeeCouncil/core/pull/298), then I can already start extending it to cover those other cases.

internetti:
> > This doesn't seem to be working for the individual form in the app, this one's still flagging a duplicate for two empty full name fields
> 
> I'll take a look. What fields have you chosen for deduplication? If it's only full name, then the issue is likely what we discussed in the thread above. But if you agree with this as the preferred approach (over #298), then I can already start extending it to cover those other cases.

It was admittedly only full name, but I just tested again with combinations of full name, mothers name and email, all with the same result

dfigueroa87:
@internetti I think it's working now, please give it another look. It's following this logic:
- If ALL fields in the criteria are empty, two entries are not considered as the same.
- If some fields in the criteria are empty, those fields are ignored and the non-empty ones considered for deduplication.

FYI, while testing I also found another issue with deduplication based on birth date (the only field which is not a string) already happening in DEV env. I created a [bug ticket](https://norwegianrefugeecouncil.atlassian.net/browse/CPT-720) for it.

internal/db/individual.go
```diff
@@ -200,14 +202,15 @@ func buildInsertIndividualsQuery(tempTableName string, schema []DBColumn, df dat
...
 	FROM individual_registrations ir
 	CROSS JOIN temp_individuals_<requestId> ti
 	WHERE ir.country_id = $1
 		AND ir.deleted_at IS NULL
-		AND (ti.birth_date = ir.birth_date)
-		AND (ti.email_1 = ir.email_1 OR ti.email_2 = ir.email_2 OR ti.email_3 = ir.email_3)
-		AND (ti.first_name = ir.first_name AND ti.middle_name = ir.middle_name AND ti.last_name = ir.last_name AND ti.native_name = ir.native_name);
-
+		AND ((ti.birth_date = ir.birth_date)
+		AND ((ti.email_1 != '' AND ti.email_1 = ir.email_1) OR (ti.email_2 != '' AND ti.email_2 = ir.email_2) OR (ti.email_3 != '' AND ti.email_3 = ir.email_3) OR (ti.email_1 = '' AND ti.email_2 = '' AND ti.email_3 =''))
```
internetti, line 211:
```suggestion
		AND ((ti.email_1 != '' AND ti.email_1 = ir.email_1) OR (ti.email_2 != '' AND ti.email_2 = ir.email_2) OR (ti.email_3 != '' AND ti.email_3 = ir.email_3) OR (ti.email_1 != '' AND ti.email_2 != '' AND ti.email_3 !=''))
```
do you mean this?

dfigueroa87: No, it seems counter-intuitive, but it's correct as it is. Since the main logical operator is `AND`, when all values (in this case emails) are empty we want that clause to be `true` so the other clauses are considered. (If it would be false, then the whole thing would become false)
And to cover the case in which all the values for every considered column are empty, there's the extra clause in line 213 checking that not everything is empty.

internetti (changes requested):
deduplicating on fullname and mothersname, All -> 2 people with empty fullnames but same mothernames are flagged as duplicates.
Based on conversations I had with Fix when I did this, these should not be flagged

Also, deduplicating on names, only giving a firstname that matches another participant's firstname (who has non-empty values for the others), they are flagged as a duplicate.
names should only be flagged if all the non-empty values are equal

dfigueroa87:
> deduplicating on fullname and mothersname, All -> 2 people with empty fullnames but same mothernames are flagged as duplicates. Based on conversations I had with Fix when I did this, these should not be flagged

This is in line with the implementation I made. Can you point me to the requirements for this feature, so I can take a look and adapt it?


> Also, deduplicating on names, only giving a firstname that matches another participant's firstname (who has non-empty values for the others), they are flagged as a duplicate. names should only be flagged if all the non-empty values are equal

Can you give me more details for this one? I couldn't reproduce it. I tried adding a new participant having the same first name as an existing one (the existing one also had the middle and last name defined, whereas in the new one I left all other fields empty), and it was added successfully.

internetti:
> > deduplicating on fullname and mothersname, All -> 2 people with empty fullnames but same mothernames are flagged as duplicates. Based on conversations I had with Fix when I did this, these should not be flagged
> 
> This is in line with the implementation I made. Can you point me to the requirements for this feature, so I can take a look and adapt it?

Difficult, this was on teams, sometime last spring ... the easiest way might be for you to talk to fix. in the end he's the one who decides how this should work


> > Also, deduplicating on names, only giving a firstname that matches another participant's firstname (who has non-empty values for the others), they are flagged as a duplicate. names should only be flagged if all the non-empty values are equal
> 
> Can you give me more details for this one? I couldn't reproduce it. I tried adding a new participant having the same first name as an existing one (the existing one also had the middle and last name defined, whereas in the new one I left all other fields empty), and it was added successfully.

... now I also can't reproduce it anymore

dfigueroa87:
> So nevermind, that would be okay then, sorry for the false alarm (although specifically in this case they could just be siblings ...)
In my opinion that'd be a misuse of the system, deduplication on those fields shouldn't be selected if adding that type of data is the expected outcome.
But anyway, I can review this together with Fix as part of the Product Review process, after merging the changes.

> ... now I also can't reproduce it anymore
Magic. My code is full of it. :joy:

internetti (commented):
Screenshot from 2024-02-15 09-51-56
Screenshot from 2024-02-15 09-52-40


Now I get this. some participants in the db without any names, same as the file, deduplicating only on names, these are flagged. also intentional?

internetti (approved):

PR #300, opened 2024-02-01
fix(deduplication): CPT-718 Compare email and phone fields among each other

dfigueroa87:
https://norwegianrefugeecouncil.atlassian.net/browse/CPT-718

For deduplication, make a cross check among the 3 different email/phone fields.
This solves the issue where adding a participant with a first email being the same as the second email of an existing participant was not being marked as a duplicate.

Note this PR is based on the changes improving the empty string comparison implemented as queries in the deduplication config, as it extends that same approach. If we instead go with the other option, the implementation needs to be moved to a generic function that would build the proper query.


neb42 (approved):

internetti (commented):
needs rebasing.

also, this might (later) also have to be added for deduplicating on IDs

dfigueroa87:
@internetti rebased :slightly_smiling_face:

internetti (approved):

PR #307, opened 2024-02-26
CPT-711: sql trigger to prevent overrides by empty values

internetti:
adding an sql trigger that checks on every update if there is an existing value that would be overwritten by an empty value, empty string or NULL, throws an exeption with some extra data.

Potential issue: this will only ever alert on one value at a time, not collect all potential overwrites from a file or form.

Comment: not convinced that this is actually wanted behaviour for all fields.


dfigueroa87:
My understanding is that in this scenario the entry should be updated, but just that the empty field would be ignored and the existing value kept. Not an error returned.

Also, my suggestion would be to avoid using triggers or adding any logic directly to the database (or only use that as a last resort approach). The reason is that this type of logic gets quite hidden and is not easy to maintain or understand what is really going on.

PR #312, opened 2024-04-10
Core24 169 moving consent up on the form

neb42:
Jira ticket: JIRA-CORE24-169

## What have you implemented?

* Add consent section to form
* Move consent questions there
* Move consent questions to start of file

## How to test it?
Look at the page


internal/views/individual_form.go
```diff
@@ -13,6 +13,7 @@ import (
 type IndividualForm struct {
 	*forms.Form
 	individual             *api.Individual
+	consentSection				 *forms.FormSection
```
dfigueroa87, line 16: Indentation is off

dfigueroa87 (commented):
Looks good, just one comment

internal/views/individual_form.go
```diff
@@ -255,6 +257,17 @@ func (f *IndividualForm) buildTitle(t locales.Translator) error {
...
 }
 
+func (f *IndividualForm) buildConsentSection(t locales.Translator) error {
+	f.consentSection = &forms.FormSection{
+		Title:       t("consent"),
+		Fields:      []forms.Field{},
+		Collapsible: true,
+		Collapsed:   false,
+	}
+	f.Form.Sections = append(f.Form.Sections, f.consentSection)
```
internetti, line 267: have you spoken to masa? ticket doesn't mention creating a new section

internal/views/individual_form.go
neb42: Where? It looks ok to me

internal/views/individual_form.go
neb42: I checked with Zineb (as her name was on the ticket), but you're right that this likely came from Masa so I'll double check with her.

PR #314, opened 2024-04-10
Make service comment field bigger

neb42:
Jira ticket: JIRA-CORE24-157

## What have you implemented?
I couldn't reproduce the issue on the ticket so I've just made the textarea bigger


dfigueroa87 (approved):

internetti (commented):
the limit is currently 65,535 characters, postgres allows to increase that to 1GB, we could also use LONGTEXT (4GB)

PR #317, opened 2024-04-29
Performance improvements

neb42:
Jira ticket: JIRA-CORE24-173

## What have you implemented?

* For languages and nationalities, don't generate the select field options server side (causing 3-4mb html file). Instead load 2 json files, cache them, and generate select options client side.
* Remove unused js files causing bloat
* Add a success alert when saving a participant to provide feedback to the user. Previously it wasn't clear that anything had happened.

## How to test it?

* Throttle network and load individual form. Compare with deployed instance and look at payload size in network tab.


dfigueroa87 (approved):

web/templates/individual.gohtml
```diff
@@ -52,6 +52,8 @@
             } else if (firstFocusableInput != null) {
                 focusInput(firstFocusableInput)
             }
+            
+            window.history.pushState({}, null, window.location.href.split('?')[0]);
```
internetti, line 56: this deletes all form data when you refresh the page after a deduplication error

internetti (approved):

PR #318, opened 2024-05-22
Increase connection max lifetime to 20 min to help with long file upl…

neb42:
## What have you implemented?

* Increase the postgres connection max lifetime from 5 mins to 20 mins. This is to help with errors caused by long uploads with deduplication. 

## Extra/next steps

* The performance issue seems to be caused by the deduplication within the file. I'm looking into improvements there but this pr should alleviate immediate problems.


dfigueroa87 (approved):
:clock1:

PR #320, opened 2024-05-30
Core24 342 deduplication performance issues

neb42:
Jira ticket: JIRA-CPT-342

## What have you implemented?

* Ensure checkbox labels are clickable in deduplication modal
* Fix deduplication check for ids so that they compare across all fields
* Fix deduplication not-empty check
* Move file deduplication to the database, using the temporary table with a cross-join on itself
* Add indices to the temporary table used in deduplication

## How to test it?

* Fully test deduplication EXTENSIVELY please, not just the performance issue
* To reproduce the performance issue, upload a 10k file and select several deduplication options. I think this will only be apparent in a deployed instance though.

## Extra/next steps

* This pr removes the need for the dataframe package. I have left it in as the error message formatting still uses it. I felt that it wasn't worth the time spent on v1 to do it.


neb42:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=NorwegianRefugeeCouncil_core&pullRequest=320) **Quality Gate failed**
> Failed conditions  [4.4% Duplication on New Code](https://sonarcloud.io/component_measures?id=NorwegianRefugeeCouncil_core&pullRequest=320&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=NorwegianRefugeeCouncil_core&pullRequest=320)

I'm ignoring this

internetti (commented):
there are some undefined things in the tests, so they're not building
docker run --rm -it -v /home/irena/Documents/code/core-v1/reports:/app/reports core-dev test
Running tests...
# github.com/nrc-no/notcore/internal/db [github.com/nrc-no/notcore/internal/db.test]
internal/db/individual_test.go:138:19: undefined: buildInsertIndividualsQuery
internal/db/individual_test.go:217:13: undefined: buildDeduplicationQuery

internetti:
> @internetti The address issue is a problem on main and touches a different bit of the code to this pr. I'll leave it out of this pr.
> 
> It also seems to be the first column, not the address, being used.

Can you create a ticket for that then?

internal/handlers/individual.go
```diff
@@ -175,38 +170,38 @@ func HandleIndividual(renderer Renderer, repo db.IndividualRepo) http.Handler {
...
 					return
 				}
-				duplicates, err = repo.FindDuplicates(ctx, df, deduplicationConfig)
-				if err != nil {
+
+				if duplicatesInFile != nil {
 					alerts = append(alerts, alert.Alert{
 						Type:        bootstrap.StyleDanger,
-						Title:       fmt.Sprintf("Error during deduplication %e", err),
+						Title:       fmt.Sprintf("Error during deduplication"),
```
internetti, line 177: I would match this with the message in line 198

internal/db/individual.go
```diff
@@ -32,46 +30,142 @@ type IndividualRepo interface {
...
+		deduplicationTempTableConfig.schema,
+	)
+	err := tx.SelectContext(ctx, &ret, deduplicationQuery, selectedCountryID)
+	if err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
+
+func findFileDuplicates(ctx context.Context, tx *sqlx.Tx, deduplicationTempTableConfig *DeduplicationTempTableConfig, config deduplication.DeduplicationConfig, selectedCountryID string, individualCount int) ([]containers.Set[int], error) {
```
internetti, line 109: selectedCountry is not needed anymore

internal/handlers/individuals_upload.go
```diff
@@ -157,33 +157,37 @@ func HandleUpload(renderer Renderer, individualRepo db.IndividualRepo) http.Hand
...
 				return
 			}
 
-			dbDuplicationErrors := api.FormatDbDeduplicationErrors(duplicatesInDB, df, deduplicationConfig)
-			if len(dbDuplicationErrors) > 0 {
-				ids := []string{}
-				for _, d := range duplicatesInDB {
-					ids = append(ids, fmt.Sprintf("id=%s", d.ID))
+			if duplicatesInFile != nil {
+				errors := api.FormatFileDeduplicationErrors(duplicatesInFile, deduplicationConfig, records, colMapping)
```
internetti, line 168: are the records still needed?

neb42: I haven't changed the error formatting so I would assume so

internal/handlers/individual.go
neb42: No, this is an actual error. On the route we deduplicate a single individual. If the file check returns a duplicate for this single person then it would be an error.

internetti (commented):
deduplicating on Names OR Fullname (Or IDs), I would have expected row 5 and 7 to be flagged because of the fullname, was not though.
only deduplicating on fullname flags it.
[generated_NamesOrFullname.csv](https://github.com/user-attachments/files/15585464/generated_NamesOrFullname.csv)
NamesOrFullname

neb42:
> deduplicating on Names OR Fullname (Or IDs), I would have expected row 5 and 7 to be flagged because of the fullname, was not though. only deduplicating on fullname flags it. [generated_NamesOrFullname.csv](https://github.com/user-attachments/files/15585464/generated_NamesOrFullname.csv) NamesOrFullname

This is an issue with the error format function and is the same on main. I will look into it.

internetti (approved):
only note I have is this case:
deduplicating on Ids And Names, when both have no ids, it's still flagged as a duplicate, but I think that is okay, and will probably be discussed when we get to the empty values ticket

PR #321, opened 2024-05-30
Update go.mod

christhai:
Jira ticket: JIRA-CPT-#


neb42:
I'd rather leave this a go 1.18. This project is mostly in maintenance mode and updating go versions can be a pain. I don't see a benefit to this.

